Issue #12, opened 2018-03-16
AzureData -> AzureCore (refactor): Remove/replace all print() (and similar) statements


NateRickard:
Once logging is implemented in AzureCore, all instances of print(), printStackTrace(), etc. must be removed/updated to use appropriate logging through AzureCore.

PR #25, opened 2018-04-05
First cut at using kotlin-logging instead of println

### LINKED ISSUE #5: AzureCore Library Dependencies: Logging

johnnylambada:

Partially Fixes #5 and #12 .

- use https://github.com/MicroUtils/kotlin-logging as a front end to the Android `Log` class
- Add a `ContextProvider.verbose` method to lazily apply the log if `verboseLogging` is true
- Use the combination in the AzureData library wherever there was a `print()` or `println()`

If this looks good then I'll apply this technique across the rest of the libraries.

Note that the `logger` object is defined separately in each class so that it picks up that classes name as the log message's `tag`.

### DIFF
--- a/azuredata/build.gradle
+++ b/azuredata/build.gradle
@@ -33,6 +33,9 @@ dependencies {
     implementation group: 'com.google.code.gson', name: 'gson', version: "$gsonVersion"
     implementation "com.squareup.okhttp3:okhttp:$okhttpVersion"
     implementation "com.squareup.okhttp3:logging-interceptor:$okhttpVersion"
+    implementation 'io.github.microutils:kotlin-logging:1.4.9'
+    // https://mvnrepository.com/artifact/org.slf4j/slf4j-android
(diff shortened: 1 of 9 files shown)


johnnylambada:
`kotlin-logging` relies on `slf4j` which in turn relies on `slf4j-android`. Older versions of `slf4j-android` seem to work, but the newest versions use the android `Log.isLoggable` api to filter logs. It looks like this API is almost universally ignored and pretty broken. `Log.isLoggable` blocks all calls to `Log.d` (debug) and `Log.v` (verbose) on both my phone and the emulator I tested. There's supposed to be a way to work around it, but that would require a developer to use `adb ssh setprop` for _every_ tag they want to allow -- which is ridiculous -- and even that doesn't seem to work on the latest versions of Android. Unfortunately, I can't find the primary source code to `slf4j-android`, or I'd try pulling this out. When I try to just compile against the older version, the build fails because both versions are added.

PR #26, opened 2018-04-05
First cut at using timber instead of println

### LINKED ISSUE #5: AzureCore Library Dependencies: Logging

johnnylambada:
Partially Fixes #5 and #12 .

- remove `verboseLogging` from AzureData. Instead use the facilities provided by the logging library.
- use [timberkt](https://github.com/ajalt/timberkt) as a front end to the Android Log class. timberkt is a front end to the [Timber](https://github.com/JakeWharton/timber)  library.
- Use timberkt wherever there is a print() or println()

### DIFF
--- a/azurecore/build.gradle
+++ b/azurecore/build.gradle
@@ -31,7 +31,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
     implementation "com.squareup.okhttp3:okhttp:$okhttpVersion"
     implementation group: 'com.google.code.gson', name: 'gson', version: "$gsonVersion"
-    implementation 'com.jakewharton.timber:timber:$timberVersion'
+    implementation "com.jakewharton.timber:timber:$timberVersion"
     // https://mvnrepository.com/artifact/commons-codec/commons-codec
     implementation 'commons-codec:commons-codec:1.11'
 }
--- a/azurecore/build.gradle
+++ b/azurecore/build.gradle
@@ -31,6 +31,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
     implementation "com.squareup.okhttp3:okhttp:$okhttpVersion"
     implementation group: 'com.google.code.gson', name: 'gson', version: "$gsonVersion"
+    implementation 'com.jakewharton.timber:timber:$timberVersion'
     // https://mvnrepository.com/artifact/commons-codec/commons-codec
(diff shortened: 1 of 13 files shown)


colbylwilliams:
Thanks @johnnylambada - few comments/questions:

1. Is it possible (and does it make sense) to add the dependency on timberkt _only_ to azurecore and expose the logging functionality via a static class/singleton in azurecore?  This is the approach we took in the iOS library so if we ever decide to switch out the functionality (or realize we need to do something like "scrub" PPI) we only have to do it in a single place.  Actually, the scenario of insuring PPI/IDs get "scrubbed" and the possibility of adding "remote" logging in the future were main reasons we decided on implementing a unified logging lib, so I think to take full advantage we'd need it to live exclusively in azurecore.  ...so instead of importing `com.github.ajalt.timberkt.d` in azuredata, we'd import something like `com.azure.core.logging.d` and  _that_ logging file would utilize timberkt...  @johnnylambada @NateRickard can you weigh in?
2. Don't worry about replacing `print()` or` println()` within the unit/integration tests, only needs to be implemented in the actual source libraries.

NateRickard:
Yes I agree we should abstract the logging stuff away in core and use a singleton/static (companion object in kotlin) so we're just importing the logging package from core.

Also, we'd only want to enable `setPrettyPrinting` on Gson in debug/verbose mode.  If @colbylwilliams wants the verbose logging flag gone we will have to only enable pretty printing in a debug compilation.

colbylwilliams:
For the `setPrettyPrinting` @NateRickard @johnnylambada you can take advantage of the fact timberkt uses closures/blocks that'll only execute if the corresponding logging level is active. so you could do something like this (btw, I don't know kotlin):

```kotlin
val gson: Gson =
        GsonBuilder()
                .disableHtmlEscaping()
                .setConditionalPrettyPrinting()
                .registerTypeAdapter(Date::class.java, DateTypeAdapter())
                .registerTypeAdapter(Timestamp::class.java, TimestampAdapter())
                .registerTypeAdapter(DictionaryDocument::class.java, DocumentAdapter())
                .create()

fun GsonBuilder.setConditionalPrettyPrinting() : GsonBuilder {  

    logger.d {
        this.setPrettyPrinting()   
    }

    return this
}
```

johnnylambada:
@colbylwilliams ah -- I like it. It will produce a log message if logging is enabled, so it'll probably be better if it was something like this:

```
logger.d {
        this.setPrettyPrinting()   
        return "Gson.setPrettyPrinting enabled"
    }
```

PR #35, opened 2018-04-17
Reachability 

### LINKED ISSUE #4: AzureCore Library Dependencies: Reachability

johnnylambada:
Fixes #4 .

- Use ReactiveNetwork for network availability


johnnylambada:
Ok, regarding the changes you mentioned:
1. I've created the com.azure.core.network.Connectivity class to encapsulate ReactiveNetwork. For now it can only report when the network connectivity has changed, but we can add more functionality as we need it.
2. I've moved `isOffline` to `ContextProvider` since `isOffline` requires an Android `Context` to do its job. I also moved `ContextProvider` to AzureCore since it'll be used by more than AzureData.
3. Once we're comfortable with the above changes, I'll implement the `isOffline` everywhere.

Issue #14, opened 2018-03-16
Implement support for Pagination of query results and feeds


NateRickard:
Once design is complete (issue #13 ). This issue is a placeholder for implementing that design (will likely be broken out into several additional issues).

johnnylambada:
Closed with https://github.com/Azure/Azure.Android/pull/36

PR #36, opened 2018-04-19
Implement Paging of results

### LINKED ISSUE #13: Design support for Pagination of query results and feeds
[Pagination of query results](https://docs.microsoft.com/en-us/rest/api/documentdb/querying-documentdb-resources-using-the-rest-api#pagination-of-query-results)

johnnylambada:

Fixes #13 & #14.

- Added maxPerPage in all the functions that support returning a list of results
- Implemented adding the x-ms-max-item-count header
- Implemented the ListResponse<T>.next extension.
- Tests


johnnylambada:
@NateRickard I have paging working for Document types. Right now it works like this:

        var waitForResponse : ListResponse<TDoc>? = null
        createNewDocuments(3)

        // Get the first one
        AzureData.getDocuments(collectionId, databaseId, docType, 1) { waitForResponse = it }
        await().until { waitForResponse != null }
        waitForResponse.let {
            // Do something with the first page
        }

        // Get the second one
        waitForResponse.let { response ->
            waitForResponse = null
            AzureData.nextDocuments(response!!, docType) {
                // Do something with the next page
            }
        }

I'd like to get away from `AzureData.nextDocuments(response!!, docType) {}` and just do `response.next() {}` for all document types (databases, collections, documents, etc). 

I started down that road first, but it was too confusing to bite that off first, so I went down this simpler road. Now that it's working I'm going to try to do as I suggest above. If that doesn't work out, then I'll create `AzureData.nextDatabases`, `AzureData.nextCollections`, etc.

johnnylambada:
@NateRickard ok, you can get the next page using the response. It's only slightly more awkward than iOS :). The `<TDoc>` is the type of object we're dealing with. I couldn't find a way to get rid of that.

        // Get the next page
        waitForResponse.let { response ->
            waitForResponse = null
            response!!.next<TDoc> {
                // Do something with it (this page)
            }
        }
        await().until { waitForResponse != null }

Also, there is some type casting shenanigans going on [here](https://github.com/Azure/Azure.Android/blob/ad353c23e7cf0c489a1b648f88ea3a9576cf301c/azuredata/src/main/java/com/azure/data/service/Response.kt#L77-L79). They should be safe but the compiler warns.

I'd like to add more tests for each of the other Resources that can be returned in a list, but I can do that later if you rather I move on to something else.

azuredata/src/main/java/com/azure/data/service/Response.kt
```diff
@@ -21,7 +19,13 @@ open class Response<T>(
...
         val jsonData: String? = null,
         // The result of response deserialization.
-        val result: Result<T>
+        val result: Result<T>,
+        // The resourceLocation, filled out when there could be more results
+        val resourceLocation: ResourceLocation? = null,
+        // The class of the document
+        val documentClass: Class<*>? = null,
+        // The DocumentClient used to acquire this Response
+        val documentClient: DocumentClient? = null
```
NateRickard: Rather than passing in the `DocumentClient` here, let's use the AzureData 'static' calls since it will already have its doc client instance.  We can make the methods internal if it makes sense.  This should avoid the need to store the doc client in the response here and also support other, non Doc DB, data stores which should also support paging.

azuredata/src/main/java/com/azure/data/service/Response.kt
```diff
@@ -21,7 +19,13 @@ open class Response<T>(
         //  The json data returned by the server (if applicable)
         val jsonData: String? = null,
         // The result of response deserialization.
-        val result: Result<T>
+        val result: Result<T>,
+        // The resourceLocation, filled out when there could be more results
+        val resourceLocation: ResourceLocation? = null,
+        // The class of the document
+        val documentClass: Class<*>? = null,
```
NateRickard: Let's rename the param since this will support more resource types than docs.

azuredata/src/main/java/com/azure/data/service/Response.kt
```diff
@@ -58,4 +62,22 @@ open class Response<T>(
...
+    fun <T : Resource> next(callback: (ListResponse<T>) -> Unit) {
+        val documentClient = documentClient
+                ?: return callback(ListResponse(DataError(DocumentClientError.NextCalledTooEarlyError)))
+
+        val listResponse : ListResponse<T> = this as ListResponse<T>
+
+        val documentClass: Class<T> = documentClass as Class<T>
+
+        documentClient.next(listResponse, documentClass, callback)
+    }
```
NateRickard: Let's see if this is possible to implement as an extension to `ListResponse`... not sure if that works on a typealias, but would love for the API to only show up on things where it makes sense (queries, lists)

azuredata/src/main/java/com/azure/data/service/Response.kt
johnnylambada: The reason I chose to store the documentClient was if the user had created their own client, I thought it would be important to re-use that one instead of using the static one. Can the user ever build their own client and use it instead of using the static one in `AzureData`?

PR #54, opened 2018-07-29
Add a custom deserializer for ResourceList

### LINKED ISSUE #48: Investigate and fix an issue with the deserialization of ResourceList<DictionaryDocument>
## Bug Information
The key-value pairs in the `data` property of `DictionaryDocument` are missing when the `DictionaryDocument` is embedded in a `ResourceList`. The key-value pairs are available in the JSON but missing from the deserialized `ResourceList`. Most probably, there is an issue with the deserialization.

ftchirou:
Fixes #48.

This PR fixes an issue with the deserialization of `ResourceList<T>` when a custom type adapter has been registered with `Gson` for `T`. When `T` has a custom type adapter, the type adapter is not used during the deserialization of `ResourceList<T>`.

This PR fixes it by "manually" deserializing objects of type `ResourceList<T>` by first converting the JSON into `JsonElement`s and then reconstructing the objects from the JSON nodes.

### DIFF
--- a/azuredata/src/main/java/com/azure/data/service/DocumentClient.kt
+++ b/azuredata/src/main/java/com/azure/data/service/DocumentClient.kt
@@ -15,6 +15,7 @@ import com.azure.data.constants.MSHttpHeader
 import com.azure.data.model.*
 import com.azure.data.model.indexing.IndexingPolicy
 import com.azure.data.util.*
+import com.azure.data.util.json.ResourceListJsonDeserializer
 import com.azure.data.util.json.gson
 import com.google.gson.reflect.TypeToken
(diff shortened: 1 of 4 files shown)


NateRickard:
@ftchirou I had to update this branch from master to merge but that broke the build :(  Can you review the build log and see what's going on there?  Also, as noted above, a test covering the use case this PR is fixing would be great.

ftchirou:
Hi @NateRickard! I'll take a look at the build. About the test case, there are already test cases for this bug (all the `list` tests for `DictionaryDocument`); those were failing, that's how I noticed this bug.

Issue #94, opened 2019-03-07
createDocument, updateDocument, storedProcedure operations fail due to missing partition key header


SogoGolf:
hi @NateRickard , i came across the same partition key error today in our android app for:
```
.createDocument(xxx...)
and
.replaceDocument(xxx...)
and
.executeStoredProcedure(xxx...)
```
error:
```
{"code":"BadRequest","message":"Message: {\"Errors\":[\"The partition key supplied in x-ms-partitionkey header has fewer components than defined in the the collection.\"]}
```
would you be able to apply the same fix made for the ```.deleteDocument``` ? Much appreciated.

NateRickard:
@SogoGolf Can you confirm that the `@PartitionKey` annotation on your model class(es) resolved this issue for you?  If so let's close this... still interested in the PR you mentioned as I can see a situation where passing the value directly would be needed.

SogoGolf:
hi @NateRickard using the annotation tested ok for both createDocument and updateDocument. 

for the executeStoredProcedure ->

this was still throwing the partition key error but let me explain what im trying to do --> I'm calling the stored proc, shaping the data in the stored proc, then returning the payload. the client side object that i'd want to deserialise to, doesn't have/need the partition key field if that makes sense. here's the details:

__model__
```
data class ClubsNearMe(
        val id: String,
        val name: String,
        val metersFromMe: Float
){
    class Deserializer: ResponseDeserializable<Array<ClubsNearMe>> {
        override fun deserialize(content: String): Array<ClubsNearMe>? = Gson().fromJson(content, Array<ClubsNearMe>::class.java)
    }
}
```

im using ```Fuel``` http lib to call an Azure Function for the time being until i work out how to successfully call the stored proc's and get data back. with Fuel, the return payload is being automatically deserialised by Fuel to the model above.

__http call__
```
Fuel.post("https://zzz-js.xxxxxx.net/api/postQuery").jsonBody(query).responseObject(ClubsNearMe.Deserializer()) { request, response, result ->

            val (clubsNearMe, err) = result
etc....
}
```

__stored proc__
```
function getClubsNearMe(prefix) {
    var collection = getContext().getCollection();

    var isAccepted = collection.queryDocuments(
        collection.getSelfLink(),
        `SELECT f.name, ST_DISTANCE(f.location, {'type': 'Point', 'coordinates':[151.2357330322266, -33.77001515278013]}) as metersFromMe FROM etc..") < 50000`,

[...]

PR #95, opened 2019-03-13
partition keys on createDocument/updateDocument/executeStoredProcedure

SogoGolf:
Apologies didn't have time to review all the codebase's structure but it seemed that the partition key scaffolding was already in place. Even if there is more to do hopefully this pull request will help. 

added partitionKey param to createDocument
added partitionKey param to updateDocument
added partitionKey param to executeStoredProcedure

### DIFF
--- a/azuredata/src/main/java/com/azure/data/AzureData.kt
+++ b/azuredata/src/main/java/com/azure/data/AzureData.kt
@@ -231,24 +231,48 @@ class AzureData {
 
         // create
         @JvmStatic
+        @Deprecated("Creating a document without a partition key is deprecated and will be removed in a future version of AzureData")
         fun <T : Document> createDocument(document: T, collectionId: String, databaseId: String, callback: (Response<T>) -> Unit) =
                 documentClient.createDocument(document, collectionId, databaseId, callback)
 
         // create
         @JvmStatic
+        @Deprecated("Creating a document without a partition key is deprecated and will be removed in a future version of AzureData")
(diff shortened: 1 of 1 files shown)


NateRickard:
I started implementing some of these changes, but it appears the partition key value MUST exist on the document definition itself, otherwise the server will throw an error:

`{"code":"BadRequest","message":"Partition key provided either doesn't correspond to definition in the collection or doesn't match partition key field values specified in the document.\r\nActivityId: 6b838609-72e2-4cf0-b80e-835d941d2bea, \r\nRequestStartTime: 2019-03-15T21:18:16.2480577Z, RequestEndTime: 2019-03-15T21:18:16.2480577Z, Number of regions attempted: 1\r\n, Microsoft.Azure.Documents.Common/2.2.0.0"}`

So, since you're defining the partition key as a _property_ on the collection's documents (e.g. `/testKey`), the server will validate that this property exists AND matches the header value sent.

As mentioned in #94, `execute` is another story and that's been updated in 0.2.2 so you can pass whatever value you need to.

NateRickard:
On second thought, I did go ahead and add the overloads to createDocument() and createOrUpdateDocument() to support manually passing in the partition key value.  I can see this being needed if you don't own the model to add the annotation so the sdk can automatically extract the value.  Note that you would still need to have the property on the document with the matching key value set.  Added in #98 since this PR was incomplete and I already added the execute() support.

PR #118, opened 2019-11-13
Initializing structure for android sdk & azure-core

anuchandy:


sdk/build.gradle
```diff
@@ -0,0 +1,29 @@
+// Top-level build file where you can add configuration options common to all sub-projects/modules.
+
+buildscript {
+    repositories {
+        google()
+        jcenter()
+    }
+
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.4.2'
```
vcolin7: Structure looks good. Just as a note, there is a newer stable Gradle version: 3.5.2

sdk/core/azure-core/build.gradle
```diff
@@ -0,0 +1,38 @@
+apply plugin: 'com.android.library'
+
+android {
+    compileSdkVersion 29
+    buildToolsVersion "29.0.1"
```
vcolin7: Newer version of this as well: 29.0.2

anuchandy:
@vcolin7 good catch on L21, will update the PR

sdk/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Wed Nov 13 14:53:27 PST 2019
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-all.zip
```
bsiegel: Gradle itself is also up to v6.0. You can regenerate the gradle wrapper with:

```
./gradlew wrapper --gradle-version=6.0
```

An exciting new feature (to me anyhow!) is **dependency version alignment**:
> **Aligning versions across multiple dependencies**
> 
> Dependency version alignment allows builds to express that different modules belong to the same logical group (like a platform) and need to have identical (a.k.a aligned) versions in a dependency graph.

bsiegel: Gradle 6.0 supports the Android Gradle plugin v3.4 and up, so we should be good to go there. There is one known issue where Gradle 6.0 seems to have broken incremental compilation for Android, but this will be fixed in 6.0.1.

anuchandy: `gradle wrapper --gradle-version=6.0.1` worked.

PR #124, opened 2019-11-18
test service for sas generation

anuchandy:
A small test-service that can be hosted locally on dev-machine to test externalizing-sas-generation scenario when we build iOS/Android TransferManager. It can 

(1). create containers and 
(2). generate  SaS token for blobs i.e. if we provide container name and name of the file to be uploaded then SaS for that file will be generated.


bsiegel:
Depending on how serious we want to make the code review for this tool, my suggestion would be to make the endpoints more RESTful - an example could be:

PUT /containerName => create container
GET /containerName/blobName/token?duration => generate sas token

If this were our main codebase I'd lobby for something like the above, but since this is just a tool, it's probably just a nice to have.

anuchandy:
@bsiegel thanks for looking into this. One reason we used POST for SaS generation was - Azure generally follows the pattern of using POST for generating credentials, one example is [here](https://github.com/Azure/azure-rest-api-specs/blob/0a7f5e5079a81337de3995e71b6e5ee4dbf48756/specification/storage/resource-manager/Microsoft.Storage/stable/2019-06-01/storage.json#L417).

The being said the main reason we used POST was - POST can be used with "Fixed Endpoint" like `\create-container`, `\generate-sas`.  I wanted to avoid parsing of "Arbitrary Endpoint" like `/{containerName}`, `/{containerName}/{blobName}/token?{duration}`, since handling such generic endpoint requires more work in controller level.

May be we should keep it simple now as this is just a tool we use for dev purpose, without spending much cycles on making it production ready.  Thoughts?

bsiegel:
> May be we should keep it simple now as this is just a tool we use for dev purpose, without spending much cycles on making it production ready. Thoughts?

Totally fine with that, if it was super easy I feel like it'd be worthwhile but this is just a dev tool, so it's not worth spending more than a trivial amount of effort on.

bsiegel:
One remaining suggestion would be to move this under either `tools/` which is (sort of) a convention that our other repos seem to have already adopted.

bsiegel:
:+1::sparkles:Ship It:sparkles::shipit:

Issue #127, opened 2019-11-20
Create logger interface and implement HTTP Pipeline logging policies


anuchandy:
We've a ClientLogger contract defined in Java azure-core, we should be able to reuse most of it. It's pluggable for SLF4J adapters. There exists an slf4j-android `org.slf4j:slf4j-android` which by default forward log calls to https://developer.android.com/reference/android/util/Log.html

  - [x] Logger interface
  - [x] HTTP logger policy
  - [x] cURL-formatted logger policy

PR #128, opened 2019-11-21
Azure Core request interceptors

### LINKED ISSUE #133: Implement HTTP Pipeline request modification policies
Create interceptors for policies that modify the outgoing request:
  - [x] User agent policy
  - [x] Request ID policy
  - [x] Add date policy

For the user agent, we want to report:

application ID / bundle ID
library name and version
os.version
user.region
user.language
min & target api level / os version
built for debugging
device model & manufacturer
language (kotlin vs java)

Determine which of these metrics we can access, then propose an update to the guidelines with the format of the useragent string that we will use to report them. Implement the useragent string to match that guideline.

Also note: The telemetry on-off toggle was not intended to affect the value of the User-Agent header. We don't need to implement any logic inside this policy related to checking the value of that toggle.

vcolin7:

Fixes #133

- Adding dependency on ThreeTenABP for use of time APIs
- Added OkHttp Interceptors for requests in Azure Core


vcolin7:
Created draft PR to get feedback from the team :)

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/AddDateInterceptor.java
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.core.http.interceptors;
+
+import org.jetbrains.annotations.NotNull;
```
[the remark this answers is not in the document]
anuchandy: agree, its my bad that I pulled it in when we created the PR for repo struct.

vcolin7:
@JonathanGiles @anuchandy Changed the JetBrains dependency to one from AndroidX. It's lightweight and having these type of annotations can prove useful.

From the [documentation](https://developer.android.com/studio/write/annotations#adding-annotations): "If you use annotations in your own library module, the annotations are included as part of the Android Archive (AAR) artifact in XML format in the annotations.zip file. Adding the support-annotations dependency does not introduce a dependency for any downstream users of your library."

vcolin7:
Left our logging initialization and calls until that work is finalized (issue #127).

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
```diff
@@ -0,0 +1,152 @@
...
+     * @param version {@link ServiceVersion} of the service to be used when making requests.
+     * @param configuration Configuration store that will be checked for the AZURE_TELEMETRY_DISABLED.
+     * @throws NullPointerException if {@code configuration} is {@code null}.
+     */
+    public UserAgentInterceptor(String applicationId, String sdkName, String sdkVersion, Configuration configuration,
+                           ServiceVersion version) {
+        Objects.requireNonNull(configuration, "'configuration' cannot be null.");
+        boolean telemetryDisabled = configuration.get(Configuration.PROPERTY_AZURE_TELEMETRY_DISABLED, false);
+        if (telemetryDisabled) {
+            this.userAgent = String.format(DISABLED_TELEMETRY_USER_AGENT_FORMAT, sdkName, sdkVersion,
```
vcolin7: Is there any reason for using one extra argument (`version.getVersion()`) at the end of the formatting call? All statements where we set the `userAgent` through the use `String.format()` have this additional element as a parameter.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/CoreUtils.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /*/**
+     * Turns an array into a string mapping each element to a string and delimits them using a coma.
+     * @param array Array being formatted to a string.
+     * @param mapper Function that maps each element to a string.
+     * @param <T> Generic representing the type of the array.
+     * @return Array with each element mapped and delimited, otherwise null if the array is empty or null.
+     * /
+    public static <T> String arrayToString(T[] array, Function<T, String> mapper) {
```
vcolin7: Left this method out since it makes use of the Java 8 Stream API and it's not used by any of the interceptors in this PR. Will address this at a later stage.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/CoreUtils.java
```diff
@@ -0,0 +1,177 @@
...
+    }*/
+
+
+    /*/**
+     * Helper method that returns an immutable {@link Map} of properties defined in {@code propertiesFileName}.
+     *
+     * @param propertiesFileName The file name defining the properties.
+     * @return an immutable {@link Map}.
+     * /
+    public static Map<String, String> getProperties(String propertiesFileName) {
```
vcolin7: Left this method out since it makes use of the Java 8 Stream API and it's not used by any of the interceptors in this PR. Will address this at a later stage.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/CoreUtils.java
```diff
@@ -0,0 +1,177 @@
...
+
+    /*/**
+     * Extracts and combines the generic items from all the pages linked together.
+     * @param page The paged response from server holding generic items.
+     * @param context Metadata that is passed into the function that fetches the items from the next page.
+     * @param content The function which fetches items from the next page.
+     * @param <T> The type of the item being returned in the paged response.
+     * @return The publisher holding all the generic items combined.
+     * /
+    public static <T> Publisher<T> extractAndFetch(PagedResponse<T> page, Context context,
```
vcolin7: Left this method out since it makes use of Reactor and it's not used by any of the interceptors in this PR. Will address this at a later stage.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/CoreUtils.java
vcolin7: If we do not want to rewrite stuff that involves the Stream APIs, there is a couple libraries that may be worth considering, specially the last one mentioned in this [StackOverflow answer](https://stackoverflow.com/a/39515397/4967034): android-retrostreams. However, by reading the documentation I noticed it uses a GPLv2 license. I know this means we cannot look at the source code but I'm fairly certain that we could add ask authorization to add a dependency on it, I'm just not exactly sure of how this whole process works. Do you know something about this @adrianhall?

bsiegel: Anything that we might expose that uses Java 8 Streams APIs wouldn't necessarily be something an Android developer would expect. I'd be fine just leaving them out. For this particular method, there are lots of other ways to accomplish the same task, so if we need it I would just write it without using Streams.

bsiegel: There are a bunch of retro* projects that let you use various modern Java features on Android, but I'd shy away from using any of them. Many app developers don't want to use them, and as time goes on, more and more of them are being rendered unnecessary due to the desugaring support in Android Studio.

adrianhall: 1) Our core position should be that every dependency should be idiomatic to Android (i.e. not forcing Android developers into uncommon patterns), common in the ecosystem, and provide significant value. 
2) I'm not sure streams is idiomatic to android anyhow.

We probably want to look at Azure Core as "only implement what is needed to implement the core service library we need" rather than "implement a copy of azure-core from somewhere else"

vcolin7:
Cleaned up the PR and removed methods and classes that are not necessary for the current interceptors.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/AddDateInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Locale;
+
+import okhttp3.Interceptor;
+import okhttp3.Response;
+
+/**
+ * The pipeline interceptor that adds a "Date" header in RFC 1123 format when sending an HTTP request.
+ */
+public class AddDateInterceptor implements Interceptor {
+    private final DateTimeFormatter format = DateTimeFormatter
```
bsiegel: My spidey senses are telling me that we should make this static, it's going to get reused over and over - and while the interceptor instances themselves are pretty long-lived, each client will have at least one.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/RequestIdInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+public class RequestIdInterceptor implements Interceptor {
+    private static final String REQUEST_ID_HEADER = "x-ms-client-request-id";
+
+    @NonNull
+    @Override
+    public Response intercept(@NonNull Chain chain) throws IOException {
+        Request request = chain.request();
+        String requestId = request.header(REQUEST_ID_HEADER);
+
+        if (requestId == null)
```
bsiegel: Let's add the missing curly braces here, it's a stylistic thing I know but this is one I will always argue for.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/function/Function.java
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.core.util.function;
+
+/**
+ * Simple interface make code compatible with Java 7
+ */
+public interface Function<Input, Output> {
```
bsiegel: We should not duplicate the Java 8 interface types if possible. I know that Android doesn't contain the function types, but since the Android toolchain can compile code with Lambdas, there has to be some Type or Interface it uses to represent them. Can we see what type that is?

sdk/core/azure-core/src/main/java/com/azure/android/core/util/CoreUtils.java
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.core.util;
+
+/**
+ * This class contains utility methods useful for building client libraries.
+ */
+public class CoreUtils {
```
bsiegel: Should mark this `final` as well. OR, what about using an interface with static methods (since the Android toolchain does now support these on all API levels)?

sdk/build.gradle
```diff
@@ -20,10 +20,12 @@ allprojects {
     }
 
     ext {
+        annotationsVersion = "28.0.0"
```
bsiegel: Same comment here as in 139 about the dependency version.

sdk/core/azure-core/build.gradle
```diff
@@ -34,5 +34,7 @@ android {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation 'androidx.annotation:annotation:1.1.0'
```
bsiegel: Same comment here as in 139 about the dependency version. Also we should be consistent about double vs single quotes

sdk/core/azure-core/src/main/java/com/azure/android/core/util/Configuration.java
```diff
@@ -0,0 +1,318 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
bsiegel: We probably need to get together with Anu and determine how to re-implement this. There's no way to set system properties or environment variables on Android so that doesn't apply.

The platform native way would be via [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences) which are accessible via the `Context`. However we don't want to couple this directly to the Context itself, and as we discussed a bit previously we should access any platform services or anything contained within the Context by way of an adapter interface or provider. That will allow us to write unit tests for this without having any dependencies on the Emulator.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
```diff
@@ -0,0 +1,152 @@
...
+import okhttp3.Response;
+
+/**
+ * Pipeline interceptor that adds "User-Agent" header to a request.
+ * <p>
+ * The format for the "User-Agent" string is outlined in
+ * <a href="https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy">Azure Core: Telemetry policy</a>.
+ */
+public class UserAgentInterceptor implements Interceptor {
+    private static final String DEFAULT_USER_AGENT_HEADER = "azsdk-java";
```
bsiegel: Probably needs to change, my guess is `azsdk-android` but I would check with Adrian.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
```diff
@@ -0,0 +1,152 @@
...
+
+    // From the design guidelines, the default user agent header format is:
+    // azsdk-java-<client_lib>/<sdk_version> <platform_info>
+    private static final String USER_AGENT_FORMAT = DEFAULT_USER_AGENT_HEADER + "-%s/%s (%s)";
```
bsiegel: This differers from the Java implementation (no parentheses are used there). Any reason for the change?

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
```diff
@@ -0,0 +1,152 @@
...
+    // From the design guidelines, the application id user agent header format is:
+    // AzCopy/10.0.4-Preview azsdk-java-<client_lib>/<sdk_version> <platform_info>
+    private static final String APPLICATION_ID_USER_AGENT_FORMAT = "%s " + USER_AGENT_FORMAT;
+
+    // When the AZURE_TELEMETRY_DISABLED configuration is true remove the <platform_info> portion of the user agent.
+    private static final String DISABLED_TELEMETRY_USER_AGENT_FORMAT = DEFAULT_USER_AGENT_HEADER + "-%s/%s";
+
+    // From the design guidelines, the platform info format is:
+    // <language runtime>; <os name> <os version>
+    private static final String PLATFORM_INFO_FORMAT = "%s; %s %s";
```
bsiegel: This probably needs to be a change in our design guidelines. The runtime is always known, and the "OS" name and version probably don't make sense, instead we probably want to get some info about the Android API level / version, the OEM, etc. This also seems like an Adrian discussion.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
```diff
@@ -0,0 +1,152 @@
...
+     * @param applicationId User specified application Id.
+     * @param sdkName       Name of the client library.
+     * @param sdkVersion    Version of the client library.
+     * @param version       {@link ServiceVersion} of the service to be used when making requests.
+     * @param configuration Configuration store that will be checked for the AZURE_TELEMETRY_DISABLED.
+     * @throws NullPointerException if {@code configuration} is {@code null}.
+     */
+    public UserAgentInterceptor(String applicationId, String sdkName, String sdkVersion, Configuration configuration,
+                                ServiceVersion version) {
+        Objects.requireNonNull(configuration, "'configuration' cannot be null.");
```
bsiegel: Can we just use @NonNull annotation in the constructor parameters and avoid the assert here?

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
```diff
@@ -0,0 +1,152 @@
...
+     * in the user agent.
+     *
+     * @param applicationId User specified application Id.
+     * @param sdkName       Name of the client library.
+     * @param sdkVersion    Version of the client library.
+     * @param version       {@link ServiceVersion} of the service to be used when making requests.
+     * @param configuration Configuration store that will be checked for the AZURE_TELEMETRY_DISABLED.
+     * @throws NullPointerException if {@code configuration} is {@code null}.
+     */
+    public UserAgentInterceptor(String applicationId, String sdkName, String sdkVersion, Configuration configuration,
```
bsiegel: We should be able to read the application ID from the Context via `Context.getPackageName()`, however again we don't want to couple ourselves directly to Context so we should discuss how to create an abstraction around it.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
```diff
@@ -0,0 +1,152 @@
...
+     * User-Agent header is updated by prepending the value in this interceptor.
+     * {@inheritDoc}
+     */
+    @NonNull
+    @Override
+    public Response intercept(@NonNull Chain chain) throws IOException {
+        Request request = chain.request();
+        String header = request.header("User-Agent");
+
+        if (header == null || header.contains(DEFAULT_USER_AGENT_HEADER)) {
```
bsiegel: This differs from the Java implementation, there the check is `header.startsWith` rather than `contains`. Any reason for the change?

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
```diff
@@ -0,0 +1,152 @@
...
+            header = userAgent;
+        } else {
+            header = userAgent + " " + header;
+        }
+
+        return chain.proceed(request
+            .newBuilder()
+            .addHeader("User-Agent", header)
```
bsiegel: This should be `.header()` except for circumstances where the header is multiply-valued (rare, e.g. Cookie)

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/AddDateInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+        .ofPattern("EEE, dd MMM yyyy HH:mm:ss 'GMT'")
+        .withZone(ZoneId.of("UTC"))
+        .withLocale(Locale.US);
+
+    @NonNull
+    @Override
+    public Response intercept(@NonNull Chain chain) throws IOException {
+        return chain.proceed(chain.request()
+            .newBuilder()
+            .addHeader("Date", format.format(OffsetDateTime.now()))
```
bsiegel: This should be `.header()` except for circumstances where the header is multiply-valued (rare, e.g. Cookie)

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/RequestIdInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+            request = request.newBuilder()
+                .addHeader(REQUEST_ID_HEADER, UUID.randomUUID().toString())
```
bsiegel: This should be `.header()` except for circumstances where the header is multiply-valued (rare, e.g. Cookie)

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
```diff
@@ -0,0 +1,152 @@
...
+            .build());
+    }
+
+    private static String getPlatformInfo() {
```
bsiegel: I don't believe any of this applies / makes sense on Android.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/function/Function.java
vcolin7: Did some research and it seems that the AndroidX has a [package](https://developer.android.com/reference/androidx/arch/core/util/package-summary) for this.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/CoreUtils.java
vcolin7: Changing it to an interface with static methods after our in-person discussion.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/function/Function.java
vcolin7: According to our discussion today, we will have a deeper discussion about the how we access platform data, which will involve the Configuration class (the only one using this interface so far). Depending on how this goes we might not even need to rely on AndroidX.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
vcolin7: Will have a discussion about this with the team during our next stand-up meeting.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/AddDateInterceptor.java
adrianhall: Should the interceptors be public API surface?  I'm thinking it's likely an internal thing for right now.  We want to expose as little as possible, and increase the surface area as we need to.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/AddDateInterceptor.java
adrianhall: Dumb point, but I'm thinking long term maintainability - can we name the format variable something like 'httpDateFormatter'?

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
vcolin7: Will be removing these for now. A deeper discussion about how to handle telemetry on mobile needs to be had though.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/UserAgentInterceptor.java
vcolin7: The Java repo has been recently updated, I'll look into the changes.

vcolin7:
Hey @bsiegel & @anuchandy, I made changes to the UserAgentInterceptor to include the metrics mentioned in issue #133. There are a couple of things that are pending some discussion or investigation on my part, but I wanted to have something you could review for now.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/UserAgentInterceptor.java
```diff
@@ -3,13 +3,19 @@
 
 package com.azure.android.core.http.interceptor;
 
+import android.content.Context;
```
bsiegel: Whenever we need to reference a class in the `android` package from within some business logic (in this case, the user agent interceptor), I've found it extremely beneficial to create an interface to decouple them. If we don't decouple them, we're unable to unit-test the UserAgentInterceptor. So instead I might imagine:

```java
package com.azure.android.core.provider;
public interface AppMetadataProvider {
    public String getApplicationVersion();
    public String getTargetSdkVersion();
    ...
}

---

package com.azure.android.core.provider.impl;
public class AndroidAppMetadataProvider implements AppMetadataProvider {
    public AndroidAppMetadataProvider(Context context) {
        this.appInfo = context.getApplicationInfo();
        ...
    }

    public String getTargetSdkVersion() {
        return Integer.toString(this.appInfo.targetSdkVersion);
    }

    ...
}

---

package com.azure.android.core.http.interceptor;
public class UserAgentInterceptor implements Interceptor {
    public UserAgentInterceptor(AppMetadataProvider provider, String userAgent) {
        ...
    }
}
```

Using DI here to decouple the Android system APIs lets us easily write unit tests by defining and passing in a `TestAppMetadatProvider` for unit tests that can return stub values.

bsiegel: To clarify, the example above is the simple case, where we only need to avoid calling Platform API methods. If the API methods also return an instance of a Platform Type, we would need to wrap that type as well.

vcolin7: Thank you for the thorough explanation! Will add these changes later this week.

vcolin7:
Hey @bsiegel, I created a couple interfaces and their implementation to decouple from some the Android platform types. There are some pending TODOs I'd like to have a discussion about though :)

vcolin7:
.

PR #139, opened 2019-11-23
Azure Core response interceptors

### LINKED ISSUE #136: Implement HTTP Pipeline response policies
Create interceptors for policies that operate on the response:
  - [x] Response validation policy
  - [x] Scrub eTag policy

vcolin7:

Fixes #136.

- Added OkHttp Interceptors for responses in Azure Core


sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,6 @@
+package com.azure.android.core.util.logging;
+
+public class ClientLogger {
```
vcolin7: Created stub for my project to compile. This class is a part of PR #129 by @anuchandy.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
```diff
@@ -0,0 +1,45 @@
...
+import com.azure.android.core.util.logging.ClientLogger;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+
+import okhttp3.Interceptor;
+import okhttp3.Response;
+
+public class ResponseValidationInterceptor implements Interceptor {
```
vcolin7: Created only the interceptor class, no builder like in the [Java SDK](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-common/src/main/java/com/azure/storage/common/policy/ResponseValidationPolicyBuilder.java) for now. Maybe @JonathanGiles or @alzimmermsft could shed some light into why the Java team decided to take the builder route.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
```diff
@@ -0,0 +1,45 @@
+package com.azure.android.core.http.interceptors;
```
bsiegel: This interceptor is unique to Storage only, so we should put it in that package.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ScrubEtagInterceptor.java
```diff
@@ -0,0 +1,40 @@
+package com.azure.android.core.http.interceptors;
```
bsiegel: This interceptor is unique to Storage only, so we should put it in that package.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
bsiegel: Let's redesign this a little bit. I see that in Java, the only functionality exposed / used is the builder's addOptionalEcho method, in order to verify a request header's value matches the same header's value in the response. We should be able to simplify this and move from a generic "response validator" to single, concrete "response header echo validator" or something (I'm sure we can come up with a better name). If there are other types of validations we want to add in the future, they can be added as different interceptors.

So my suggestion would be to make an interceptor with a constructor that can accept a list of header names, and whose `intercept` implementation matches the logic that `ResponseValidationPolicyBuilder.addOptionalEcho` currently performs in the Java SDK.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ScrubEtagInterceptor.java
```diff
@@ -0,0 +1,40 @@
...
+
+import okhttp3.Interceptor;
+import okhttp3.Response;
+
+/**
+ * Wraps any potential error responses from the service and applies post processing of the response's eTag header to
+ * standardize the value.
+ */
+public class ScrubEtagInterceptor implements Interceptor {
+    private static final String ETAG = "ETag";
```
bsiegel: This differs from the Java implementation, there they use the value `eTag`. Any reason for the change?

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ScrubEtagInterceptor.java
```diff
@@ -0,0 +1,40 @@
...
+        Response response = chain.proceed(chain.request());
+        String eTag = response.header(ETAG);
+
+        if (eTag == null) {
+            return response;
+        }
+
+        eTag = eTag.replace("\"", "");
+
+        return response.newBuilder()
```
bsiegel: Two questions:

First, is this the only way to modify the headers? It seems inefficient but maybe it's not and that's just OkHTTP's design to always use a builder to make any modification to the request or response. I think this is the performance concern Adrian had mentioned.

Second, is the `header()` method case-insensitive? That is, if we retrieve the value from a header named "etag" and then replace it here using the name "ETag" will it work as expected or will it end up adding a second header?

sdk/core/azure-core/src/main/java/com/azure/android/core/util/function/BiConsumer.java
```diff
@@ -0,0 +1,8 @@
+package com.azure.android.core.util.function;
+
+/**
+ * Simple interface make code compatible with Java 7
+ */
+public interface BiConsumer<Input1, Input2> {
```
bsiegel: If we end up taking my suggestion above about de-abstracting the Response Validation interceptor, we can get rid of this class.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
bsiegel: Let's just make sure to remove this before merging.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ScrubEtagInterceptor.java
vcolin7: I'm not sure why the Java SDK is using "eTag", but since the HTTP [specification](https://tools.ietf.org/html/rfc7232#section-2.3) states that the name of the header is "ETag" and all examples I've seen online use it as well, I decided to stick to this casing.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
vcolin7: So, summarizing, instead of having a class that can have a list of assertions to check the response against, we would be talking about an interceptor that validates that the values of a given list of headers are consistent between the request and the response?

sdk/build.gradle
```diff
@@ -20,10 +20,11 @@ allprojects {
     }
 
     ext {
+        annotationsVersion = "28.0.0"
```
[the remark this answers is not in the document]
vcolin7: Agreed, I think I missed that. The correct and most updated version of said library at this time is [1.1.0](https://mvnrepository.com/artifact/androidx.annotation/annotation/1.1.0).

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ScrubEtagInterceptor.java
vcolin7: According to OkHttp's [documentation](https://square.github.io/okhttp/interceptors/#rewriting-requests), this is the way of modifying any headers on a request/response.

vcolin7: As for the `header()` method, let me do some research on this.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
bsiegel: Right, because that is the only type of assertion we ever coded into the Java SDK. It seems like the Java SDK implemented it as a way to allow us to write any arbitrary type of assertions, but only one type of assertion has ever actually been created so far. I don't see the value in adding that complexity for Android.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ScrubEtagInterceptor.java
bsiegel: I'm surprised that OkHTTP doesn't seem to have a standardized enum or class that contains the header names. So I'm good with this change.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
vcolin7: In the Java SDK, the `RequestValidationPolicy` class is public, doesn't this mean that it would also allow users to add their own assertions?

bsiegel:
Also just a small nit, but we should be consistent with quote marks in our gradle files. We seem to use both double and single quotes more or less at random. I don't care which one wins, though for consistency with Java I think I prefer double quotes.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ScrubEtagInterceptor.java
vcolin7: Just checked and it seems that when you try to get a header value the case does not matter, however, I would still prefer to follow the specification.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ScrubEtagInterceptor.java
vcolin7: This method is case-insensitive.

vcolin7:
> 
> 
> Also just a small nit, but we should be consistent with quote marks in our gradle files. We seem to use both double and single quotes more or less at random. I don't care which one wins, though for consistency with Java I think I prefer double quotes.

By default, Android Studio uses single quotes. We use double quotes when we want to replace a variable like this: `implementation "com.squareup.okhttp3:okhttp:$okHttpVersion"`

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
bsiegel: I would assume so, but I don't personally feel like there's a compelling reason to expose that feature on mobile. However the buck stops with @adrianhall on that :)

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
vcolin7: Should we move it in the future or right now? I'm not sure if I recall correctly but I think we said we'd have these in core for now and move them once we started working on storage.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ScrubEtagInterceptor.java
vcolin7: See above comment

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
bsiegel: I don't think I was there for that discussion. I would probably move them to the storage package now (creating any missing project structure).

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
anuchandy: Agree with @bsiegel, we can keep it either as a package internal type or move it under implementation package. Storage specifically want to validate two headers `x-ms-client-id` and `x-ms-encryption-key-sha256`. If customers are interested in custom validations they can provide their own interceptors.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptors/ResponseValidationInterceptor.java
anuchandy: It was me suggested putting them in core and moving out eventually when we start do storage. Agree, lets create storage-package, which is needed anyway if we want to have these to be package internal.

vcolin7:
> 
> 
> Also just a small nit, but we should be consistent with quote marks in our gradle files. We seem to use both double and single quotes more or less at random. I don't care which one wins, though for consistency with Java I think I prefer double quotes.

I played around with Android Studio a bit and I found that both singe and double quotes are used in auto-generated gradle files. For the sake of consistency, I agree with using only double quotes form now on since we need them anyways to use variables in our strings.

PR #140, opened 2019-11-25
Adding ClientLogger

### LINKED ISSUE #127: Create logger interface and implement HTTP Pipeline logging policies

anuchandy:
Adding logging abstraction. Partially fixes #127, this PR address only " Logger interface" sub item.

  - [ ] Logger interface
  - [ ] HTTP logger policy
  - [ ] cURL-formatted logger policy

The last two will be a separate PR.


bsiegel:
@anuchandy I see this PR has the client logger interface, but it doesn't complete the associated issue - we're missing the two logging policies. Did you plan to add those in this PR as well? Or in a separate PR?

anuchandy:
@bsiegel The logging policies will be a seperate PR, I thought of getting this in first so that other PR want to use logging infrastructures can be unblocked.

I've updated the PR description with this info.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,289 @@
...
+        if (doesArgsHaveThrowable(args)) {
+            // If we are logging an exception the format string is already the exception message, don't append it.
+            if (!isExceptionLogging) {
+                Object throwable = args[args.length - 1];
+
+                // This is true from before but is needed to appease SpotBugs.
+                if (throwable instanceof Throwable) {
+                    throwableMessage = ((Throwable) throwable).getMessage();
+                }
+            }
```
vcolin7: In the Java SDK there's a [piece of code](https://github.com/Azure/azure-sdk-for-java/blob/a2fa92b55d4c82000ccc64c135efbe1a9dd45746/sdk/core/azure-core/src/main/java/com/azure/core/util/logging/ClientLogger.java#L201) that removes the throwable from he args array at the end of this block. I think it would be valuable to keep it.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,289 @@
...
+    private Object[] removeThrowable(Object... args) {
+        return Arrays.copyOf(args, args.length - 1);
+    }
+
+    /**
+     * Checks if the character sequence is null or empty.
+     * @param charSequence Character sequence being checked for nullness or emptiness.
+     * @return True if the character sequence is null or empty, false otherwise.
+     */
+    private static boolean isNullOrEmpty(CharSequence charSequence) {
```
vcolin7: I'm including a CoreUtils interface with static methods in PR #139 that includes this method, I think we can remove it from here.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,289 @@
...
+        return args[args.length - 1] instanceof Throwable;
+    }
+
+    /*
+     * Removes the last element from the arguments as it is a throwable.
+     *
+     * @param args The arguments passed to format the log message.
+     * @return The arguments with the last element removed.
+     */
+    private Object[] removeThrowable(Object... args) {
```
vcolin7: Not being used as the piece of code that calls this is missing (see previous comment).

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,289 @@
...
+     *
+     * @param format The formattable message to log.
+     * @param args Arguments for the message. If an exception is being logged, the last argument should be the
+     *     {@link Throwable}.
+     */
+    public void error(String format, Object... args) {
+        log(ERROR_LEVEL, format, args);
+    }
+
+    /*
```
vcolin7: Javadoc blocks should start with /**

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,289 @@
...
+        Objects.requireNonNull(runtimeException, "'runtimeException' cannot be null.");
+
+        if (canLogAtLevel(logLevel)) {
+            performLogging(logLevel, true, runtimeException.getMessage(), runtimeException);
+        }
+
+        return runtimeException;
+    }
+
+    /*
```
vcolin7: Javadoc blocks should start with /**

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,289 @@
...
+                }
+                logger.error(format, args);
+                break;
+            default:
+                // Don't do anything, this state shouldn't be possible.
+                break;
+        }
+    }
+
+    /*
```
vcolin7: Javadoc blocks should start with /**

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,289 @@
...
+            case WARNING_LEVEL:
+                return logger.isWarnEnabled();
+            case ERROR_LEVEL:
+                return logger.isErrorEnabled();
+            default:
+                return false;
+        }
+    }
+
+    /*
```
vcolin7: Javadoc blocks should start with /**

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,289 @@
...
+     */
+    private boolean doesArgsHaveThrowable(Object... args) {
+        if (args.length == 0) {
+            return false;
+        }
+
+        return args[args.length - 1] instanceof Throwable;
+    }
+
+    /*
```
vcolin7: Javadoc blocks should start with /**

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,289 @@
...
+ * </ol>
+ *
+ */
+public class ClientLogger {
+    private final Logger logger;
+
+    /*
+     * Indicates that log level is at verbose level.
+     */
+    private static final int VERBOSE_LEVEL = 1;
```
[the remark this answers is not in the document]
bsiegel: In Android, people tend to avoid using enums in favor of static int values with conventional names. My understanding is that the reasons have primarily to do with memory use (enums are objects), performance (enums used to be really slow on the older Dalvik runtime), and ease of use in IPC (passing value types when calling other services via Binder is much easier because you can just add them directly to the parcel - sending objects in a parcel is much more complicated and inefficient).

However, we should be using @IntDef / @StringDef annotations to provide enum-like validation checking at compile time though. See https://developer.android.com/reference/androidx/annotation/IntDef - also highly recommend just browsing through the list of annotations provided by the AndroidX library, many of them are ones that we should be using to help our customers better understand the contract of our APIs.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    /**
+     * Logs a formattable message that uses {@code {}} as the placeholder at {@code verbose} log level.
+     *
+     * <p><strong>Code samples</strong></p>
+     *
+     * <p>Logging a message at verbose log level.</p>
+     *
+     * {@codesnippet com.azure.core.util.logging.clientlogger.verbose}
```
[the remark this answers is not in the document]
bsiegel: I don't really have a good opinion on this, not sure what is easier to maintain in the long run. I'm open to hearing arguments for/against.

PR #148, opened 2019-12-10
Azure Core logging interceptors

### LINKED ISSUE #127: Create logger interface and implement HTTP Pipeline logging policies

vcolin7:

Partially fixes #127.

- Create LoggingInterceptor and some helper artifacts
- **(PENDING)** Create interceptor for logging request in a cURL format
- Create artifacts required to parse URLs


sdk/build.gradle
```diff
@@ -6,8 +6,12 @@ buildscript {
         jcenter()
     }
 
+    ext {
```
vcolin7: These same changes are being included in one of my other PRs (#128).

sdk/core/azure-core/build.gradle
```diff
@@ -1,4 +1,4 @@
-apply plugin: 'com.android.library'
+apply plugin: "com.android.library"
```
vcolin7: These same changes are being included in one of my other PRs (#128).

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LoggingInterceptor.java
```diff
@@ -0,0 +1,327 @@
...
+    public Response intercept(@NonNull Chain chain) throws IOException {
+        Request request = chain.request();
+        Response response;
+
+        // No logging will be performed, trigger a no-op.
+        if (logDetailLevel == LogDetailLevel.NONE) {
+            return chain.proceed(request);
+        }
+
+        final ClientLogger logger = new ClientLogger(request.method() + " " + getRedactedUrl(request.url()));
```
vcolin7: In the Java SDK a context parameter called "caller-method" is logged here. In absence of a context object, for now I'm logging the HTTP method type and the URL. Need to talk to someone on the java team to figure what this message looks like.

vcolin7:
At this stage, should we consider a way of checking what the system/environment logging level is? PR #140 is not checking for this and for now I have not included any logic for this as well.

bsiegel:
> At this stage, should we consider a way of checking what the system/environment logging level is? PR #140 is not checking for this and for now I have not included any logic for this as well.

I don't believe that's a concept on Android, because similar to iOS, all logging messages are always generated (except for Debug level message in release builds). The log viewing tool (for Android, this is `logcat`) is used to filter messages at the log level desired by the user of the tool.

vcolin7:
Fixed a couple issues related to LoggingInterceptor and removed unnecessary code (URL parsing classes). Also created CurlLoggingInterceptor, LogUtils class and an additional method in CoreUtils.

Maybe we can have a chat tomorrow to go over these changes @bsiegel. What do you think?

anuchandy:
@bsiegel @vcolin7 slf4j logger used by the ClientLogger (our logger abstraction) redirect log calls to android logging system via Log.v and other variants. http://www.slf4j.org/android/

PR #153, opened 2020-01-14
Created ClientLogger interface and its default implementation

### LINKED ISSUE #127: Create logger interface and implement HTTP Pipeline logging policies

vcolin7:

Partially fixes #127.

- Create ClientLogger interface.
- Create a default implementation that uses Android logging mechanisms.


sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/LogLevel.java
```diff
@@ -0,0 +1,49 @@
...
+package com.azure.android.core.util.logging;
+
+/**
+ * Enum which represent logging levels used in Azure SDKs.
+ */
+public enum LogLevel {
+    /**
+     * Indicates that log level is at verbose level.
+     */
+    DEBUG(1),
```
anuchandy: I vaguly recall once Brandon told about IntDef, so was looking into IntDef and learned that ProGuard by default optimize enum to int, so we should be fine with enum @bsiegel ?

https://www.guardsquare.com/en/products/proguard/manual/usage/optimizations
https://stackoverflow.com/a/32236950

bsiegel: Based on the discussion in that SO post, it seems like that optimization may not work, or at least it's very difficult to get ProGuard to enable it. I'd suggest going with IntDef etc. unless we find that to be super burdensome.

anuchandy: Good catch Brandon!.  @vcolin7 please see this https://developer.android.com/reference/androidx/annotation/IntDef.html

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/AndroidClientLogger.java
```diff
@@ -0,0 +1,118 @@
...
+ * <ol>
+ * <li>{@link AndroidClientLogger#error(String) Error}</li>
+ * <li>{@link AndroidClientLogger#warning(String) Warning}</li>
+ * <li>{@link AndroidClientLogger#info(String) Info}</li>
+ * <li>{@link AndroidClientLogger#debug(String) Verbose}</li>
+ * </ol>
+ *
+ */
+public class AndroidClientLogger implements ClientLogger{
+    private final String TAG;
```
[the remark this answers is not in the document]
vcolin7: You're right, I must have missed it :)

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/LogLevel.java
vcolin7: Done :)

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogUtils.java
```diff
@@ -16,7 +19,7 @@
      * Attempts to retrieve and parse the Content-Length header into a numeric representation.
      *
      * @param headers HTTP headers that are checked for containing Content-Length.
-     * @return long value indicating the content length of the Request or Response
+     * @return Long value indicating the content length of the Request or Response
```
bsiegel:
```suggestion
     * @return The content length of the Request or Response
```
The type is encoded in the signature, let's not bother duplicating that in the documentation.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogOptions.java
```diff
@@ -146,17 +159,26 @@ public String getApplicationId() {
      */
     public LogOptions setApplicationId(final String applicationId) {
         if (!CoreUtils.isNullOrEmpty(applicationId)) {
+            RuntimeException exception;
+
             if (applicationId.length() > MAX_APPLICATION_ID_LENGTH) {
-                throw logger
-                    .logExceptionAsError(new IllegalArgumentException("'applicationId' length cannot be greater than "
-                        + MAX_APPLICATION_ID_LENGTH));
+                exception = new IllegalArgumentException("'applicationId' length cannot be greater than "
```
bsiegel: For iOS we just truncated this application ID. Should we just do the same here? I don't feel like it's worth throwing an error.

Also, regardless of whether you change this or not, we should make the limitation clear in the doc comment on L157.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogOptions.java
```diff
@@ -146,17 +159,26 @@ public String getApplicationId() {
      */
     public LogOptions setApplicationId(final String applicationId) {
```
bsiegel: Actually why is this in LogOptions?  AFAIK the application ID is used only for the user agent interceptor.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogOptions.java
bsiegel: Ignore this comment (or at least apply it to the appropriate place in UserAgentInterceptor) if we remove this from LogOptions

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogOptions.java
```diff
@@ -146,17 +159,26 @@ public String getApplicationId() {
...
+                exception = new IllegalArgumentException("'applicationId' length cannot be greater than "
+                    + MAX_APPLICATION_ID_LENGTH);
+
+                logger.error("Error found when setting Application ID", exception);
+
+                throw exception;
             } else if (applicationId.contains(" ")) {
-                throw logger
-                    .logExceptionAsError(new IllegalArgumentException("'applicationId' must not contain a space."));
+                exception = new IllegalArgumentException("'applicationId' must not contain a space.");
```
bsiegel: I'm trying to think of a better way to handle the assertions about the format of the application ID. Suggestions?

bsiegel: Ignore this comment (or at least apply it to the appropriate place in UserAgentInterceptor) if we remove this from LogOptions

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -3,287 +3,112 @@
...
-        this(clazz.getName());
-    }
-
-    /**
-     * Retrieves a logger for the passed class name using the {@link LoggerFactory}.
+     * <p><strong>Code samples</strong></p>
+     *
+     * <p>Logging a message at the debug log level.</p>
+     * <p>
+     * {@codesnippet com.azure.core.util.logging.clientLogger.info}
```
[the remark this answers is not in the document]
vcolin7: @bsiegel I brought this up the first time we reviewed the ClientLogger and we didn't really discussed it any further. I would prefer to remove all these references and add them when necessary, since no other piece of code has anything like it yet. 
Keep it simple. What do you think?

sdk/storage/azure-storage-common/src/main/java/com/azure/android/storage/common/interceptor/ResponseHeadersValidationInterceptor.java
```diff
@@ -18,27 +19,46 @@
...
     private static final String ENCRYPTION_KEY_SHA256_HEADER = "x-ms-encryption-key-sha256";
 
     private final Collection<String> headerNames = new ArrayList<>();
 
     /**
-     * Constructor that adds two mandatory headers used by Storage.
+     * Constructor that adds two mandatory headers used by Storage and uses the default {@link AndroidClientLogger}.
      */
     public ResponseHeadersValidationInterceptor() {
+        this(new AndroidClientLogger(ResponseHeadersValidationInterceptor.class));
```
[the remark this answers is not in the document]
vcolin7: Sounds good, thanks for that suggestion :)

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/AndroidClientLogger.java
```diff
@@ -0,0 +1,163 @@
...
+        this.tag = tag;
+    }
+
+    /**
+     * Returns this logger's log level.
+     *
+     * @return The log level.
+     */
+    @LogLevel
+    public int getLogLevel() {
```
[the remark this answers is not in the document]
vcolin7: The [documentation for IntDef](https://developer.android.com/reference/androidx/annotation/IntDef) shows this is the correct syntax. Here are some [examples](https://gist.github.com/utkan/bd8cc3d35881e081d9a5c297b9e21044) I found somewhere else.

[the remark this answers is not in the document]
vcolin7: Maybe what you propose also works, I haven't tried it though.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/AndroidClientLogger.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    /**
+     * Determines if the logger supports logging at the given log level.
+     *
+     * @param logLevel Logging level to validate.
+     * @return Flag indicating if the environment and logger are configured to support logging at the given log level.
+     */
+    private boolean canLogAtLevel(int logLevel) {
```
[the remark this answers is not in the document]
vcolin7: I didn't do that since this is a private method and we control who calls it, but I think it doesn't hurt to use annotations to make explicitly clear what is expected in this method.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/AndroidClientLogger.java
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.core.util.logging;
```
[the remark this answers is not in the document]
vcolin7: I don't feel strongly about either keeping it as it is of creating a new package. My only concert would be that we have an `.implementation` [package](https://github.com/Azure/azure-sdk-for-android/tree/dev/sdk/core/azure-core/src/main/java/com/azure/android/core/implementation) already which doesn't follow this principle. I remember someone raised a question about this package's purpose in one of our meetings to Jonathan but I can't honestly recall his reasoning. Would you mind shedding light on this again @JonathanGiles? :)

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/AndroidClientLogger.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Indicates that log level is at the error level.
+     */
+    public static final int ERROR = 4;
+
+    /**
+     * Indicates that no log level is set.
+     */
+    public static final int NOT_SET = 5;
```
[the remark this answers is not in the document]
vcolin7: @bsiegel This came from the Java SDK where they check if the level has not been set or is greater than even error so nothing is logged. I guess I don't really know what customers would prefer is the default behavior, but I think we can stick with INFO like you propose :)

Maybe we could ask the Java folks why they chose this route as well.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogOptions.java
vcolin7: @bsiegel This was originally taken from the Java SDK. Could it be that they wanted to include the Application ID in the logs?

vcolin7: Anyways, I will remove this from here for now and ask the Java team the reasoning behind this. If we see value in having it we can add it later.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogOptions.java
vcolin7: Aside from the User-Agent string, I'm not sure what we would want the Application ID for and how bad could it be if there is an error in the value passed on by customers. I'll ask the Java folks.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogOptions.java
vcolin7: I stated this on a different comment, but I think we should determine how undesirable it would be to have an incomplete or incorrect Application ID to determine the course of action.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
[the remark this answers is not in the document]
bsiegel: Sounds good to me, as part of march towards GA we will make sure to add snippets & snippet references. Until then I'm fine with them either being incorrect or missing.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/AndroidClientLogger.java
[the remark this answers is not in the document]
bsiegel: Oh, it sounds like the solution for "don't log anything" is to set the level to `NOT_SET` (which is surprising to me - I would expect it to be called e.g. `NONE`).

On iOS we provide a `ClientLoggers.none` which is basically an implementation of the [null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) - it's a `ClientLogger` which never actually logs anything.

I could go either way on this. I don't remember seeing other logging frameworks where you turn off logging by changing the verbosity level, that's why it surprised me. Would love to hear opinions from the peanut gallery.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/AndroidClientLogger.java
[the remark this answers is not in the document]
bsiegel: Yep, at the very least it provides some nice intellisense when so annotated.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/AndroidClientLogger.java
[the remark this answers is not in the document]
bsiegel: Yep, quite right. Looked up the spec and that's the right place - because we have to annotate the method itself, since there is a way to annotate a method parameter but no way to annotate a method return type.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogOptions.java
bsiegel: Sounds good, as far as I know it's only used in that one place. If we find a use for it in logging we can always add it back.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogOptions.java
bsiegel: Yep, agreed. On iOS we have exactly the same problem, except worse because we accept the various options in the constructor to `UserAgentInterceptor` and most users don't expect constructors to throw exceptions.

One challenge is that for iOS we populate the applicationId as the bundle identifier if the user doesn't supply one. If the bundle identifier is > 24 characters, then the user will get an exception without ever doing anything related to the applicationId.

For Android I would think we'd want a similar default (probably group ID + artifact ID) for applicationId, so even though this isn't being set in a constructor, it'd still be surprising. Maybe let's chat with @adrianhall and see if he has some guidance.

PR #155, opened 2020-01-21
Barebones pipelines for EngSys

mitchdenny:
This PR drops in the ```ci.yml``` file for ```sdk/core```. It points to a template that does nothing, but once it is merged we'll have PR builds running which I will then create separate PRs to build out the functionality of the pipelines.


eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -0,0 +1,19 @@
...
+  Artifacts: []
+  ServiceDirectory: not-specified
+  
+stages:
+  - stage: Build
+    jobs:
+    - template: ../jobs/archetype-sdk-client.yml
+      parameters:
+        ServiceDirectory: ${{parameters.ServiceDirectory}}
+        Artifacts: ${{parameters.Artifacts}}
```
bsiegel: The job template doesn't declare the parameter `Artifacts`

bsiegel:
Just had the one line that I think needs deleted, otherwise this looks good to go.

eng/pipelines/templates/stages/archetype-sdk-client.yml
mitchdenny: Doesn't need to, its only used when you want to have a default set. I would agree thought that across the mono-repos that this is inconsistently applied. I've tweaked it nonetheless :)

PR #160, opened 2020-01-22
Improvements to the Gradle build

bsiegel:


settings.gradle
```diff
@@ -0,0 +1,2 @@
+include ':sdk:core:azure-core'
+include ':sdk:storage:azure-storage-common'
```
anuchandy: any advantage in moving `settings.xml` file from `sdk` dir? With this here we have to point AndroidStudio to the root, root contains other artifacts not used for daily dev, wondering whether it affects editor experience.

sdk/core/azure-core/build.gradle
```diff
@@ -35,13 +43,10 @@ android {
 dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "androidx.annotation:annotation:$annotationsVersion"
+    implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonDatabindVersion"
+    implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-xml:$jacksonDataFormatXmlVersion"
     implementation "com.jakewharton.threetenabp:threetenabp:$threeTenAbpVersion"
     implementation "com.squareup.okhttp3:okhttp:$okHttpVersion"
     implementation "com.squareup.retrofit2:retrofit:$retrofitVersion"
-    implementation "com.jakewharton.threetenabp:threetenabp:$threeTenAbpVersion"
```
vcolin7: We still need the this dependency for the AddDateInterceptor

anuchandy: seems this was a duplicate entry, there is same in line48

settings.gradle
anuchandy: I cloned this PR and opened in AndroidStudio. Though it showed everything when opened first time, after the load/sync there exists only the two modules (core and storage-common). We're good here.

PR #173, opened 2020-01-30
 Created 'internal.provider' package for our provider implementations

vcolin7:
Created the 'internal.provider' package which holds the implementation of our provider interfaces. Also cleaned up the JavaDoc in AndroidClientLogger to no longer use '@inheritDoc', as it is not necessary.


sdk/core/azure-core/src/main/java/com/azure/android/core/provider/ApplicationInformationProvider.java
```diff
@@ -1,9 +1,26 @@
...
  * Interface for a provider that contains application information.
  */
 public interface ApplicationInformationProvider {
+    /**
+     * Creates a default {@link ApplicationInformationProvider}.
+     *
+     * @param context Android {@link Context} object to extract data from.
+     * @return A default {@link ApplicationInformationProvider}.
+     */
+    static ApplicationInformationProvider getDefault(Context context) {
```
bsiegel: By making this change, do we actually need the `AndroidApplicationInformationProvider` to be public any more? Meaning it would be package-private, but people can still get an instance of it by calling this public method.

bsiegel: And if the answer is that it can be package-private, should we then move it out of the internal namespace and directly into `com.azure.android.core.provider`?

vcolin7: We don't strictly need it to be public, I just thought it looked nicer to have them grouped like this. They can definitely be package private. Would we prefer that?

bsiegel: We should use "internal" in the namespace in order to discourage customers from using a public class that we can't make private or package-private for whatever reason.

If we can make it package-private or private, we should do so, and leave "internal" for those edge cases. I'm fine with leaving the implementations in their own namespace `com.azure.android.core.provider.implementation` and just tightening up the access level for these classes to the least-visible option we can (I assume that's package-private).

anuchandy: Consumers pass these providers to userAgentInterceptor Ctr like:

https://github.com/Azure/azure-sdk-for-android/blob/a9d9b5e4cf55e5de96021f171bb4e392aea7372f/sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/UserAgentInterceptor.java#L133-L138

 I guess other than the unit test scenarios, we/consumers are going to call `getDefault(Context|Configuration)` on each of these provider interfaces and pass it to `UserAgentInterceptor`.

I wonder it make sense to have a type that composes default providers then interceptor take that type.

```java
interface EnvironmentInformation {
   ApplicationInforationProvider getApplicationInformationProvider();
   LocaleInformationProvider getLocaleInformationProvider();
   PlatformInformationProvider getPlatformInformationProvider();

   static EnvironmentInformation getDefault(Context cxt, Configuration cfg) {
       return EnvironmentInformation() {
           @Override
           public ApplicationInformationProvider getApplicationInformationProvider() {
               return ApplicationInformationProvider.createDefault(cxt);
           }
   
           // and other providers
       }
  }
```

thoughts?

anuchandy: Agree with you guys on keeping these default implementation package-private (in the same namespace as provider). I think we should put a type with "PUBLIC" access modifier in `internal` only if that's that only option because we want that type outside package but don't want customers to use it.

anuchandy: @vcolin7 I didn't mean to hold on this PR with `EnvironmentInformation`. Let's discuss that in our next sync up to see it make sense. PR lgtm, just make sure the default provider impls are package-private.

bsiegel: @anuchandy I have wondered the same thing on iOS, but so far I haven't found it terribly cumbersome. I think most places will only need one or two providers, the UserAgent interceptor is sort of an exception.

bsiegel: @anuchandy One thing that might be nice is a convenience constructor that takes the Context - this would be pretty familiar to most Android devs. E.g.:

```java
public UserAgentInterceptor(@NonNull Context context,
                            @NonNull String sdkName, 
                            @NonNull String sdkVersion, 
                            @Nullable String applicationId) {
    this(sdkName,
         sdkVersion,
         applicationId, 
         PlatformInformationProvider.getDefault(context), 
         ApplicationInformationProvider.getDefault(context), 
         LocaleInformationProvider.getDefault(context)
    )
}
```
IIRC the convention is that the context is usually the first parameter.

anuchandy: @bsiegel sounds good, the `LocaleInformationProvider` requires Configuration instance to retrieve info it needed, so we might want to take "@NonNull Context context, @NonNull Configuration conf"

bsiegel: @anuchandy Can you not just retrieve that configuration object from the context?

```java
Configuration config = context.getResources().getConfiguration();
```

anuchandy: cool, then I agree with your proposal of ctr taking context.

vcolin7:
@bsiegel Just a note related to package visibility: if we put the provider classes inside an `provider.implementation` folder we cannot make them package-private, they must be public. To limit access I would suggest keeping them under `provider` and making them package-private.

PR #181, opened 2020-02-07
Feature/android release hookup

mitchdenny:
This PR includes changes to support the use of the Java release pipeline for Android packages. There isn't much here, just a flag that the pipeline needs to differentiate between a pipeline running out of the Java repo to work around an incompatibility that has crept in deeper in the pipeline.

This PR has an equivalent in the Java repo, as well as a change in the internal build tools repo (which would need to be landed first).

### DIFF
--- a/eng/pipelines/templates/stages/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/stages/archetype-sdk-client.yml
@@ -17,3 +17,4 @@ stages:
         DependsOn: Build
         Artifacts: ${{parameters.Artifacts}}
         ArtifactName: packages
+        VerifyVersions: false
--- a/eng/pipelines/templates/stages/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/stages/archetype-sdk-client.yml
@@ -12,7 +12,7 @@ stages:
 
   # The Prerelease and Release stages are conditioned on whether we are building a pull request and the branch.
   - ${{if and(ne(variables['Build.Reason'], 'PullRequest'), eq(variables['System.TeamProject'], 'internal'))}}:
-    - template: pipelines/stages/archetype-java-release.yml@azure-sdk-build-tools
+    - template: pipelines/stages/archetype-android-release.yml@azure-sdk-build-tools
       parameters:
         DependsOn: Build
         Artifacts: ${{parameters.Artifacts}}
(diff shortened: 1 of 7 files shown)


eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -16,4 +16,5 @@ stages:
       parameters:
         DependsOn: Build
         Artifacts: ${{parameters.Artifacts}}
-        ArtifactName: packages
\ No newline at end of file
+        ArtifactName: packages
+        VerifyVersions: 'false'
```
JimSuplizio: No quotes around the boolean value **false**

JimSuplizio:
> This PR includes changes to support the use of the Java release pipeline for Android packages. There isn't much here, just a flag that the pipeline needs to differentiate between a pipeline running out of the Java repo to work around an incompatibility that has crept in deeper in the pipeline.
/CC @weshaggard 
@mitchdenny The java pipeline is tailored towards releasing the Java SDK packages and the story is similar for net, js and python. I don't see the java pipeline becoming less tailored. If there are special considerations (or lack of) for Andriod then should it have its own definition?

mitchdenny:
> > This PR includes changes to support the use of the Java release pipeline for Android packages. There isn't much here, just a flag that the pipeline needs to differentiate between a pipeline running out of the Java repo to work around an incompatibility that has crept in deeper in the pipeline.
> > /CC @weshaggard
> > @mitchdenny The java pipeline is tailored towards releasing the Java SDK packages and the story is similar for net, js and python. I don't see the java pipeline becoming less tailored. If there are special considerations (or lack of) for Andriod then should it have its own definition?

When I first created the release templates I avoided any repo specific operations and instead of focused on processing the artifacts going through the pipeline. That is why you don't see many places in the rest of the templates where files are cloned from the "self" repo ... or at least that was originally the case.

For both Java and JavaScript release pipeline I have/had instances where repos other than our main mono-repos used the same release pipelines (qpid-proton-j-extensions for Java, and I think another one for JavaScript).

All that said, I think it might start to make sense to split them at this point in time since the PR update process is going to start resulting in some fairly ugly forking logic in the release templates. Might be able to get reuse at a lower level via using template fragments.

PR #185, opened 2020-02-19
WorkerManager based Async Blob upload

anuchandy:


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/SasTokenCredentialInterceptor.java
```diff
@@ -0,0 +1,37 @@
...
+    public SasTokenCredentialInterceptor(SasTokenCredential credential) {
+        this.credential = credential;
+    }
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        HttpUrl requestURL = chain.request().url();
+        String delimiter = (requestURL.query() != null && requestURL.query() != "") ? "&" : "?";
+        requestURL.toString();
+        String newURL = requestURL.toString() + delimiter + this.credential.getSasToken();
```
bsiegel: This could have a bug if the query string is not the last part of the request URL. For example, a URL can have a `#fragment` component.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -200,6 +425,308 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
...
+                            BlockBlobStageBlockHeaders typedHeader = deserializeHeaders(response.headers(),
+                                BlockBlobStageBlockHeaders.class);
+                            callback.onResponse(new BlockBlobsStageBlockResponse(response.raw().request(),
+                                response.code(),
+                                response.headers(),
+                                null,
+                                typedHeader));
+                        } else {
+                            String strContent = readAsString(response.body());
+                            callback.onFailure(new BlobStorageException(strContent, response.raw()));
```
bsiegel: Is this a failure if `response.isSuccessful()` is true?

bsiegel: (And apply this question for other instances of this pattern)

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/BlobUploadEntity.java
```diff
@@ -0,0 +1,137 @@
...
+ * Package private.
+ *
+ * Represents blob upload metadata for a single file upload.
+ *
+ * The Data Access Object type {@link UploadDao} exposes DB store and read methods on this model.
+ *
+ * @see TransferDatabase
+ */
+@Entity(tableName = "blobuploads")
+final class BlobUploadEntity {
```
bsiegel: We'd discussed in arch meeting about maybe making this generic and having "source" and "destination" properties. Would this PR be the best place to do that or did you want to address later?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/StorageBlobClientsMap.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Package private.
+ *
+ * A map of configured {@link StorageBlobClient} used by {@link UploadHandler} instances to
+ * make API calls.
+ */
+final class StorageBlobClientsMap {
```
bsiegel: As discussed in our arch chat, we'll want to think about this. If we're coming back from the app being closed, the client map would be empty (since we can't persist it to the database) and so the resume would (presumably) fail.

We probably need to permit multiple clients to be "registered" with the TransferManasger, and allow transfers to specify the client to use by name (with overloads to make it easy to use a single default client and not have to worry or understand about the "registry" process or naming them).

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param context the context
+     * @param storageUrl the blob storage url
+     * @param sasToken the SAS token
+     */
+    public TransferClient(Context context, String storageUrl, String sasToken) {
+        this(context,
+            new StorageBlobClient.Builder()
+                .setBlobServiceUrl(storageUrl)
+                .setCredentialInterceptor(new SasTokenCredentialInterceptor(new SasTokenCredential(sasToken)))
```
bsiegel: I may have missed it, but where/how does the customer provide the method that will be called to retrieve a new SAS token / preauthenticated URL when this token expires?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/UploadHandler.java
```diff
@@ -0,0 +1,337 @@
...
+     */
+    private void stageBlocks(List<BlockUploadEntity> blocks) {
+        for (BlockUploadEntity block : blocks) {
+            this.finalizeIfStopped();
+            Log.v(TAG, "stageBlocks(): Uploading block:" + block.blockId + threadName());
+            ServiceCall call = this.blobClient.stageBlock(this.blob.containerName,
+                this.blob.blobName,
+                block.blockId,
+                block.getBlockContent(),
+                null, new com.azure.android.core.http.Callback<Void>() {
```
bsiegel: Nit: Can we import this type?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/UploadWorker.java
```diff
@@ -0,0 +1,150 @@
...
+     * does is light weight. This method creates an {@link UploadHandler} and delegate the
+     * file upload work.
+     *
+     * @return the future representing the async upload.
+     */
+    @NonNull
+    @Override
+    public ListenableFuture<Result> startWork() {
+        Log.v(TAG, "startWork() called." + this);
+        ListenableFuture<Result> listenableFuture = CallbackToFutureAdapter.getFuture(completer -> {
```
bsiegel: Safe to use the lambda syntax here? I think so, just checking.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/BlobUploadEntity.java
bsiegel: (This comment would mostly apply to anywhere else that something is titled "Upload...")

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/SasTokenCredentialInterceptor.java
anuchandy: We can do something like this to be more defensive: 

```java
String sasToken = this.credential.getSasToken();
if (sasToken.startsWith("?")) {
    sasToken = sasToken.substring(1);
}

HttpUrl sasDummyUrl = new HttpUrl.Builder()
        .scheme("http")
        .host("localhost")
        .query(sas)
        .build();

HttpUrl.Builder newUrlBuilder = chain.request().url().newBuilder();
for (String qName : sasDummyUrl.queryParameterNames()) {
    newUrlBuilder.addEncodedQueryParameter(qName, httpUrl.queryParameter(qName));
}

Request newRequest = request
        .newBuilder()
        .url(newUrlBuilder.build())
        .build();

return chain.proceed(newRequest);
```

The fragment part is not something supported or has no meaning in storage blob/container space, so the chances of hitting a bug in the current logic is very low. The defensive logic has extra parsing and looping involved per request, so a minor concern is does it add unnecessary overhead.

Thoughts?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/UploadWorker.java
anuchandy: it's safe. Here is the [documentation](https://developer.android.com/reference/androidx/concurrent/futures/CallbackToFutureAdapter) with similar Java usage.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/UploadHandler.java
anuchandy: sure we can. Will update the PR.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/SasTokenCredentialInterceptor.java
anuchandy: We can do something like this to be more defensive: 

```java
String sasToken = this.credential.getSasToken();
if (sasToken.startsWith("?")) {
    sasToken = sasToken.substring(1);
}

HttpUrl sasDummyUrl = new HttpUrl.Builder()
        .scheme("http")
        .host("localhost")
        .query(sas)
        .build();

HttpUrl.Builder newUrlBuilder = chain.request().url().newBuilder();
for (String qName : sasDummyUrl.queryParameterNames()) {
    newUrlBuilder.addEncodedQueryParameter(qName, httpUrl.queryParameter(qName));
}

Request newRequest = request
        .newBuilder()
        .url(newUrlBuilder.build())
        .build();

return chain.proceed(newRequest);
```

The fragment part is not something supported or has no meaning in storage blob/container space, so the chances of hitting a bug in the current logic is very low. The defensive logic has extra parsing and looping involved per request, so a minor concern is does it add unnecessary overhead.

Thoughts?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/StorageBlobClientsMap.java
anuchandy: yes, repopulating the lookup map is something we need to brainstorm more in upcoming arch meetings.

I liked the design you showed in the last meeting, where we allow lookup (registration) using friendly names. I will add it as part of defining builder for the TransferMgr. 

The builder with configurations (e.g. block size, parallelism ..) + named client registration will be  a separate PR. As we talked in last sync, builder will be pushed after aggregation/download apis.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
anuchandy: your concern is correct, the `SasTokenCredential` type is only support const final SAS token as of today. Mutation of SAS token via callback is yet to define, that will be separate PR.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
anuchandy: That is correct.

While okhttp (or generally in http) treat 2xx code as successful, from storage service perspective only successful response status code is 201 for the stage API. This is described in swagger response section for the sample [stage-api](https://github.com/Azure/azure-rest-api-specs/blob/52f4a2bf17f0bbd5649473faea88bd3205e74c24/specification/storage/data-plane/Microsoft.BlobStorage/preview/2019-02-02/blob.json#L6032). 

Similarly for other apis as well.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/BlobUploadEntity.java
anuchandy: right, I'm thinking of considering it once we have full download APIs in place as well, so not planning in this PR.

anuchandy: you might have already noticed, all types (including Upload prefixed types) package internal, the only public type is `TransferClient`.

anuchandy: you might have already noticed, all types (including Upload prefixed types) are package internal, the only public type is `TransferClient`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/SasTokenCredentialInterceptor.java
bsiegel: How would you feel about:

```java
public Response intercept(Chain chain) throws IOException {
    HttpUrl requestURL = chain.request().url();
    String query = requestURL.query();
    if (!CoreUtils.isNullOrEmpty(query)) {
        query += "&";
    }
    query += this.credential.getSasToken();
    HttpUrl newURL = requestURL
        .newBuilder()
        .query(query)
        .build();
    Request newRequest = chain.request()
            .newBuilder()
            .url(newURL)
            .build();
    return chain.proceed(newRequest);
}
```

anuchandy: thank you, this looks very good. I think we just needed to add "?" trimming in SaS token. Let me incorporate it and test (specifically encoding part), if all good I will update the PR with it and the other one to import CallBack.

anuchandy: The `requestURL.query()` will give "decoded query string". 
The `this.credential.getSasToken()` will give "encoded query string".

So doing `requestURL.query(decoded query string + encoded query string)` will be a problem. The `query(..)` method expect a decoded value and does encoding internally.

It looks like either we need 

```java
HttpUrl sasDummyUrl = new HttpUrl.Builder()
        .scheme("http")
        .host("localhost")
        .query(sas)
        .build();
``` 

or we need to manually split the sas by `&` then each part by `=`

Finally use `addEncodedQueryParameter`.

bsiegel: It looks like both `HttpUrl` and `HttpUrl.Builder` have equivalent `encodedQuery()` methods, that should fix the problem right?

```java
public Response intercept(Chain chain) throws IOException {
    HttpUrl requestURL = chain.request().url();
    String encodedQuery = requestURL.encodedQuery();
    String sasToken = this.credential.getSasToken();

    if (!CoreUtils.isNullOrEmpty(encodedQuery)) {
        encodedQuery += "&";
    }
    encodedQuery += sasToken.startsWith("?")
        ? sasToken.substring(1)
        : sasToken;
    HttpUrl newURL = requestURL
        .newBuilder()
        .encodedQuery(encodedQuery)
        .build();
    Request newRequest = chain.request()
            .newBuilder()
            .url(newURL)
            .build();
    return chain.proceed(newRequest);
}
```

anuchandy: Good catch, I missed `requestURL.encodedQuery()` API. thanks Brandon!.

bsiegel: Honestly, I was 75% expecting those APis not to be exposed, forcing us to have to resort to something more verbose. I saw that there is a lot of useful, general-purpose URL-manipulation functionality present inside the `HttpUrl` class but sadly we can't use the vast majority of it because all the methods are package-private and the class itself is final.

anuchandy: right, OkHttp has many good things hidden, but as internal. I wish they publish it as a separate light weight packages (buffer, url-utils etc..).

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/UploadHandler.java
anuchandy: It seems we need this to be fully qualified. The reason is, the parent `Handler` class has a [Callback](https://developer.android.com/reference/android/os/Handler.Callback) contract in it, which take more precedence.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/SasTokenCredentialInterceptor.java
```diff
@@ -25,9 +26,23 @@ public SasTokenCredentialInterceptor(SasTokenCredential credential) {
...
+
+        String sasToken = this.credential.getSasToken();
+        // SAS token is already encoded so its safe to append it to the encoded query from source request.
+        encodedQuery += sasToken.startsWith("?")
+            ? sasToken.substring(1)
+            : sasToken;
+
+        HttpUrl newURL = requestURL
+            .newBuilder()
+            .query(encodedQuery)
```
bsiegel:
```suggestion
            .encodedQuery(encodedQuery)
```
Otherwise we are re-encoding the query string.

Issue #189, opened 2020-02-21
Increase code coverage of unit tests for Preview 1


bsiegel:
We need to write more unit tests in order to hit code coverage targets for:
- [x] azure-core
- [ ] ~~azure-identity~~
- [x] azure-storage-blob

The code coverage target is 70% for each framework.

vcolin7:
Azure Core local unit test coverage is currently sitting at 80% classes and 70% lines where applicable.

Azure Storage local unit test coverage is currently sitting at 91% classes and 69% lines where applicable.

PR #198, opened 2020-02-22
Unit testing azure-core interceptors

vcolin7:
Partially addresses #189. Adds unit tests for our current HTTP interceptors (policies) in the azure-core package.

These changes consider the fixes done in #201 & #211 to the UserAgentInterceptor, LoggingInterceptor and CurlLogginInterceptor, as well as any other classes they depend on (like LogUtils).


build.gradle
```diff
@@ -122,6 +122,7 @@ allprojects {
         retrofitVersion = "2.6.2"
         staxApiVersion = "1.0-2"
         threeTenAbpVersion = "1.2.1"
+        threeTenBpVersion = "1.4.0"
```
bsiegel: Why can't we use ThreeTenApb in the tests?

bsiegel: Thanks - that makes sense. How hard would it be to just stick with java.util.Date & co? We could make some utility functions that accept ThreeTenABP instances and convert them to work with Date. I know that API sucks but if it's not a TON of work it might be nice to avoid adding this dep (even if it's only for test).

vcolin7: I don't think it would be too hard but we have some classes like [DateTimeRfc1123](https://github.com/Azure/azure-sdk-for-android/blob/dev/sdk/core/azure-core/src/main/java/com/azure/android/core/util/DateTimeRfc1123.java) the where we still probably need to import this dependency for testing, which is what I'm doing right now on [this PR](https://github.com/Azure/azure-sdk-for-android/pull/214).

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/AddDateInterceptor.java
```diff
@@ -20,12 +20,13 @@
  * HTTP request.
  */
 public class AddDateInterceptor implements Interceptor {
-    private static final String DATE_HEADER = "Date";
```
bsiegel: We shouldn't unhide fields just so we can access them from tests. It ends up with our tests knowing too much about our internal structure of the code under test. When writing tests, it's okay to have some duplication rather than having to DRY up everything.

Issue #200, opened 2020-02-25
UserAgentInterceptor does not trim or truncate user-provided Application IDs


vcolin7:
The design guidelines specify that the Application ID should not contain spaces or be longer than 24 characters.

PR #201, opened 2020-02-25
User agent interceptor fix

### LINKED ISSUE #199: UserAgentInterceptor adds additional User-Agent header instead of adding a prefix to an already populated one
The expected behavior should be to take the populated header value and add a prefix with all the system and build information to it.

vcolin7:

Fixes #199 & #200.

- Changed the intercept method to use header() instead of addHeader() so we only have one header named User-Agent at all times.
- Added checks and logic to make sure a user-provideed Application ID does not contain spaces and is not longer than 24 characters.
- Made additional changes to ensure user-provided values sdkName and sdkVersion are not included in the User Agent string when null.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/UserAgentInterceptor.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/UserAgentInterceptor.java
@@ -107,6 +107,9 @@ public class UserAgentInterceptor implements Interceptor {
                                 PlatformInformationProvider platformInformationProvider,
                                 ApplicationInformationProvider applicationInformationProvider,
                                 LocaleInformationProvider localeInformationProvider) {
+        sdkName = sdkName == null ? "" : sdkName;
+        sdkVersion = sdkVersion == null ? "" : sdkVersion;
+
         if (applicationId == null) {
             userAgent = String.format(
                 USER_AGENT_FORMAT,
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/UserAgentInterceptor.java
```diff
@@ -24,7 +25,8 @@
  * <a href="https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy">Azure Core: Telemetry policy</a>.
  */
 public class UserAgentInterceptor implements Interceptor {
-    private static final String USER_AGENT_HEADER = "User-Agent";
+    static final String USER_AGENT_HEADER = "User-Agent";
```
vcolin7: Made package-private since the unit test class will access this field.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/UserAgentInterceptor.java
```diff
@@ -34,7 +36,7 @@
     // From the design guidelines, the application ID user agent header format is:
     // [<application_id>] azsdk-android-<client_lib>/<sdk_version> (<platform_info>; <application_info>;
     // <user_locale_info>
-    private static final String APPLICATION_ID_USER_AGENT_FORMAT = "[%s] " + USER_AGENT_FORMAT;
+    private static final String USER_AGENT_FORMAT_WITH_APPLICATION_ID = "[%s] %s";
```
vcolin7: Renamed in favor of clarity.

PR #203, opened 2020-02-26
TransferInfo LiveData 

anuchandy:
#### Features:

1. TransferInfo: a model representing current state of transfer. From transfer APIs, users will receive a `LiveData` that emits `TransferInfo` events.
2. SerialExecutor: A variant of `Executor` contract that serially delegate `TransferClient` book keeping work to another `Executor`.
3. TransferClient.Builder: Adding a minimal builder around `TransferClient`.

#### Bug Fixes

1. Fixing a bug in UploadDao w.r.t aggregating uploaded bytes.

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/SerialExecutor.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/SerialExecutor.java
@@ -1,0 +1,103 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.storage.blob.transfer;
+
+import androidx.annotation.NonNull;
+import androidx.work.WorkManager;
+
+import java.util.ArrayDeque;
+import java.util.concurrent.Executor;
+
+/**
+ * Package private.
+ *
+ * An {@link Executor} that wrap and delegate work execution to another {@link Executor}.
+ * SerialExecutor ensure that the submitted works are executed serially using
+ * the wrapped executor.
+ *
+ * SerialExecutor is used by {@link TransferClient} for all its internal book keeping,
+ * which includes creating DB entries for transfer workers, querying DB for status,
+ * submitting transfer request to {@link WorkManager}.
+ *
+ * This is similar to a {@link android.os.Looper} that serially process messages from
+ * it's message queue. The difference is while Looper hold a dedicated thread throughout
+ * it lifetime, SerialExecutor delegate the work execution to another executor.
+ *
+ * @see <a href="https://developer.android.com/reference/java/util/concurrent/Executor">Executor</a>
+ */
+final class SerialExecutor implements Executor {
+    // the queue holding the tasks to be executed serially.
+    private final ArrayDeque<Task> tasksQueue = new ArrayDeque<>();
+    // the executor to delegate the task execution to.
+    private final Executor executor;
+    // the actively executing task.
+    private volatile Task activeTask;
+
+    /**
+     * Create SerialExecutor.
+     *
+     * @param executor the executor to delegate the serial execution of tasks
+     */
+    SerialExecutor(@NonNull Executor executor) {
+        this.executor = executor;
+    }
+
+    /**
+     * Submit a work to execute.
+     *
+     * @param runnable the unit of work
+     */
+    @Override
+    public synchronized void execute(@NonNull Runnable runnable) {
+        this.tasksQueue.add(new Task(runnable, this));
+        if (this.activeTask == null) {
+            this.enqueueNextTask();
+        }
+    }
+
+    /**
+     * Enqueue the next task for execution.
+     */
+    private synchronized void enqueueNextTask() {
+        this.activeTask = this.tasksQueue.poll();
+        if (this.activeTask != null) {
+            this.executor.execute(this.activeTask);
+        }
+    }
+
+    /**
+     * A Task representing a unit of work. Once a Task is executed, it request the
+     * executor to enqueue the next Task for execution.
+     */
+    private static class Task implements Runnable {
+        // the unit of work.
+        final Runnable runnable;
+        // the executor to request execution of next task.
+        final SerialExecutor serialExecutor;
+
+        /**
+         * Create a Task.
+         *
+         *  @param runnable the unit of work this task represents
+         * @param serialExecutor the executor to enqueue the execution of next task
+         */
+        Task(Runnable runnable, SerialExecutor serialExecutor) {
+            this.runnable = runnable;
+            this.serialExecutor = serialExecutor;
+        }
+
+        /**
+         * Execute this task, once completed enqueue the next task.
+         */
+        @Override
+        public void run() {
+            try {
+                this.runnable.run();
+            } finally {
+                this.serialExecutor.enqueueNextTask();
+            }
+        }
+    }
+}
+
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
@@ -3,31 +3,46 @@
 
 package com.azure.android.storage.blob.transfer;
 
+import android.annotation.SuppressLint;
 import android.content.Context;
+import android.os.Build;
 import android.util.Log;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.RequiresApi;
+import androidx.lifecycle.LiveData;
+import androidx.lifecycle.MutableLiveData;
 import androidx.work.Constraints;
 import androidx.work.Data;
 import androidx.work.ExistingWorkPolicy;
 import androidx.work.NetworkType;
 import androidx.work.OneTimeWorkRequest;
 import androidx.work.WorkManager;
+import androidx.work.impl.WorkManagerImpl;
 
 import com.azure.android.storage.blob.StorageBlobClient;
-import com.azure.android.storage.blob.credentials.SasTokenCredential;
-import com.azure.android.storage.blob.interceptor.SasTokenCredentialInterceptor;
 
 import java.io.File;
 import java.util.List;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 
 /**
- * A type that exposes APIs for blob transfer.
+ * A type that exposes blob transfer APIs.
  */
 public class TransferClient {
     private static final String TAG = TransferClient.class.getSimpleName();
+    // the application context.
     private final Context context;
-    private final TransferDatabase db;
+    // the default storage client for all transfers.
     private final StorageBlobClient blobClient;
+    // the constraints to meet to run the transfers.
+    // currently hold the network type required for transfers.
+    private final Constraints constraints;
+    // the executor for internal book keeping.
+    private SerialExecutor serialTaskExecutor;
+    // reference to the database holding transfer entities.
+    private final TransferDatabase db;
 
     /**
      * Creates a {@link TransferClient} that uses provided {@link StorageBlobClient}
@@ -35,26 +50,18 @@ public class TransferClient {
      *
      * @param context the context
      * @param blobClient the blob storage client
+     * @param constraints the constraints to meet to run transfers
+     * @param serialTaskExecutor the executor for all internal book keeping purposes
      */
-    public TransferClient(Context context, StorageBlobClient blobClient) {
+    private TransferClient(Context context,
+                           StorageBlobClient blobClient,
+                           Constraints constraints,
+                           SerialExecutor serialTaskExecutor) {
         this.context = context;
-        this.db = TransferDatabase.get(context);
         this.blobClient = blobClient;
-    }
-
-    /**
-     * Creates a {@link TransferClient} that uses provided blob storage endpoint and SAS token.
-     *
-     * @param context the context
-     * @param storageUrl the blob storage url
-     * @param sasToken the SAS token
-     */
-    public TransferClient(Context context, String storageUrl, String sasToken) {
-        this(context,
-            new StorageBlobClient.Builder()
-                .setBlobServiceUrl(storageUrl)
-                .setCredentialInterceptor(new SasTokenCredentialInterceptor(new SasTokenCredential(sasToken)))
-                .build());
+        this.constraints = constraints;
+        this.serialTaskExecutor = serialTaskExecutor;
+        this.db = TransferDatabase.get(context);
     }
 
     /**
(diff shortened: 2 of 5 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/SerialExecutor.java
```diff
@@ -0,0 +1,103 @@
...
+ * which includes creating DB entries for transfer workers, querying DB for status,
+ * submitting transfer request to {@link WorkManager}.
+ *
+ * This is similar to a {@link android.os.Looper} that serially process messages from
+ * it's message queue. The difference is while Looper hold a dedicated thread throughout
+ * it lifetime, SerialExecutor delegate the work execution to another executor.
+ *
+ * @see <a href="https://developer.android.com/reference/java/util/concurrent/Executor">Executor</a>
+ */
+final class SerialExecutor implements Executor {
```
bsiegel: How would you feel about just using `androidx.work.impl.utils.SerialExecutor` instead? It's public, but it's also in an `impl` package. However it does pretty much exactly the same thing as this class.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -14,82 +17,189 @@
...
+                    constraintsBuilder.setRequiredNetworkType(NetworkType.METERED);
+                    break;
+                case TransferNetworkType.NOT_ROAMING:
+                    constraintsBuilder.setRequiredNetworkType(NetworkType.NOT_ROAMING);
+                    break;
+                default:
+                    throw new IllegalArgumentException("networkType must be set.");
+            }
+            if (this.serialBackgroundExecutor == null) {
+                this.serialBackgroundExecutor =  new SerialExecutor(Executors.newFixedThreadPool(2));
```
bsiegel: How would you feel about using WorkManager's `Configuration.getTaskExecutor()` here as the default? The docs say that this function "Gets the Executor used by WorkManager for all its internal business logic".

bsiegel: (We would still need to wrap it in a `SerialExecutor`)

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/SerialExecutor.java
anuchandy: The thing I'm worried about there is - using a type from their "impl" package which they can move around any time. Like we are allowed to do for public types in our "implementation|internal" namespace.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdMappedToTransferInfo.java
```diff
@@ -0,0 +1,227 @@
...
+/**
+ * Package private.
+ *
+ * A type to transform an input LiveData that emit a transfer id to an output LiveData that emit
+ * a stream of {@link TransferInfo}.
+ *
+ * Each {@link TransferInfo} event from output LiveData describes current state of the transfer
+ * identified by the transfer id.
+ */
+final class TransferIdMappedToTransferInfo {
```
bsiegel: Just to clarify, the only reason we initialize this class with a `LiveData<long>` rather than just `long` is that we're using the LiveData as a communication channel, waiting for the transfer operation to be stored in the database so we can get its ID which happens in a callback. This implies that we only ever expect the `transferIdLiveData` to update once, right? 

If my understanding is correct we should document this somewhere. It wasn't obvious until I stared at this for awhile why we were using a `LiveData<long>` at all.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
anuchandy: that sounds better, let me check task manager is accessible in app.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -14,82 +17,189 @@
...
+         * Set the {@link Executor} used by {@link TransferClient} for all its internal
+         * book keeping, which includes creating DB entries for transfer workers, querying DB
+         * for status, submitting transfer request to {@link WorkManager}.
+         *
+         * TransferClient will enqueue maximum of two command to the serialBackgroundExecutor at any time.
+         *
+         * @param executor the serialBackgroundExecutor
+         * @return Builder with provided serialBackgroundExecutor set
+         */
+        public Builder setBookKeepingExecutor(@NonNull Executor executor) {
```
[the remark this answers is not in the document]
anuchandy: cool agree, let follow the same naming convention that room and WM follows.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdMappedToTransferInfo.java
anuchandy: you're right - we use `LiveData<Long>` as a communication channel, to receive the identifier that is not available at the time of creating `TransferIdMappedToTransferInfo` instance.

Regarding once vs multi emission - it emits only once in the current code base, but I have a plan to extend it for resume, after user pause, in such cases it can emit multiple times. Yet to be designed.

I will add reason for using  `LiveData<Long>` instead of long in the java-doc.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -14,82 +17,189 @@
...
+        }
+
+        /**
+         * Sets the particular {@link TransferNetworkType} the device should be in
+         * for running the transfers.
+         *
+         * @param networkType The type of network required for transfers to run
+         * @return Builder with provided network type set
+         */
+        public Builder setRequiredNetworkType(@TransferNetworkType int networkType) {
```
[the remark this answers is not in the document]
anuchandy: I'm not opposed to taking `Constraints` instance directly. I think its good to honor additional constrains given those options make sense (like charging required, battery low).

Looks like there is no nice way to create a new `Constraints` instance from an existing one along with mutation (like `newBuilder()` on `Constraints`). Looking into it.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
anuchandy: I don't see there is a way to access the current configuration of WorkManager, though it is settable during application [bootstrap](https://developer.android.com/topic/libraries/architecture/workmanager/advanced/custom-configuration#implement-configuration-provider)

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/SerialExecutor.java
bsiegel: That's true - in practice I haven't seen that happen, but it's definitely a consideration. They wouldn't have to do a major version bump if they wanted to break compatibility with this class, and we don't want to pin to an exact version since we're a library.

I guess my suggestion is probably to keep a copy of their version in our SDK and use the appropriate license + attribution, I see that your implementation and theirs are very similar but not quite identical.

Also, I know we do the same thing (put public classes not part of our public API in an `impl` package) - should we do the same for non-public classes? It feels like we should just so that all of that non-public-API code is in the same place.

bsiegel: Also I just noticed that for the public impl classes or public methods that are not part of our public API we can add `@RestrictTo` which is an IDE hint to not use those classes / methods. We can use `LIBRARY` or `LIBRARY_GROUP` as appropriate.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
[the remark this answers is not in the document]
bsiegel: It might be odd, but maybe we can just accept a `Constraints.Builder`? There's also a public `setRequiredNetworkType` method on the configuration but it has an annotation to hint to the IDE that it's not meant to be part of their public API. We could consider suppressing that lint (though, same concerns as before - this can be brittle).

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
bsiegel: Yeah, once again I see plenty of ways to do it but only if we cask to `WorkManagerImpl`. It's so frustrating that so many improvements could be made if just a few things were not hidden from the public API.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/SerialExecutor.java
anuchandy: yes, I saw that androidx annotation used in jetpack libs. 

If we've a "customer consumable public type" on which we have any "public methods" that we don't want customers to use then we should use this annotation. So far for the types in "transfer namespace", we haven't come across any cases yet.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
anuchandy: true, I felt the same for couple of core types.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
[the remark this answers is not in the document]
anuchandy: I think we should do some exploration around, whether taking builder as a param is a pattern in community. 

We can also think of taking `Constraints` instance and create a copy of it (by going through each property, bit verbose though) with required mutation applied.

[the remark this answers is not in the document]
anuchandy: what do you think?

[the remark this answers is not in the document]
bsiegel: I thought of just making a copy, but annoyingly there are not getters for every property that can be set by the builder. The `addContentUriTrigger` and all of the `setTrigger...` methods on the builder end up setting fields of the `Constraints` object which have no public getters :(

I can't think of any use case where a customer would want to add / configure such a trigger for their transfer but if they did we would miss it. Maybe the best approach is to just lift the other four `Constraints.Builder` methods we care about into our `TransferClient.Builder`:
- setRequiresBatteryNotLow
- setRequiresCharging
- setRequiresDeviceIdle
- setRequiresStorageNotLow

Also I think I'd be fine if we changed this method to accept an `androidx.work.NetworkType`, I don't think I have a concern about leaking this type from WorkManager, we can assume customers are already reasonably familiar with this type or at least can look up the docs for it from Google. But I could be convinced either way, I just don't see a ton of value being provided by wrapping it with our own enum.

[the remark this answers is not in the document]
anuchandy: hey @bsiegel thank you for doing all these investigations and adding lot of values while I'm heads down into immediate customer issue.

Except `setRequiredNetworkType`, all other setters you listed takes primitive boolean so only type we have to consider is `androidx.work.NetworkType`. I don't have any concern in exposing `androidx.work.NetworkType` directly. Let's go with that and during user study we can watch how users adapt to it and can adjust based on feedback, if any.

If app provide `androidx.work.NetworkType.NOT_REQUIRED` then I think we can just throw an IllegalArgumentException.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/SerialExecutor.java
anuchandy: reg `SerialExecutor`: I referred the "Serial" Executor example in here as well https://developer.android.com/reference/java/util/concurrent/Executor. In a way androidx SerialExecutor is also very close to this. 

Just wondering should we wait a bit more before we include one more TPL to list, I mean wait for the cases where we really cannot avoid copy-pasting 3rd party code. Like 310 time types.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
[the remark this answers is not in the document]
bsiegel: I'm onboard with that approach 👍

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
bsiegel: Maybe it's best we just run on our own thread. But I'm wondering - if most of the bookkeeping is just interacting with Room, and room uses its own thread pool, do we actually need to have a task executor? It's not a rhetorical question - I honestly don't know. Is there much more to the bookkeeping than that?

anuchandy: The upload method is called from UI thread, even though room may have it's own pool internally, still we're not allowed to make room queries from UI thread, room api throws if it find that query is initialed from main thread.

This means, we need to offload the work to a different thread. Question is,  should we spin up a new thread per transfer API call for bookkeeping vs should we hold a thread pool to do these bookkeeping.

Spinning and tearing down short lived threads per transfer api call could add overhead, given thread creation is not cheap.

Having "our own thread pool" can save thread creation cost, but it still hold threads that is only used for bookkeeping purposes and not accessible for customers.

Having a delegated TaskExecutor, enables us to still do the bookkeeping, at the same time we leave an option for customers to share their executor with us that they also use for some their own work. We use it carefully, in the sense, we don't overwhelm it with too many works in parallel given there is a serialization.

There seems many trade-off here across options.

bsiegel: Seems so. I guess it's fine if we have our own small task executor, and if a customer wants to optimize they can create their own executor and pass it to both the configuration of WorkManager / Room and to our TransferClient.

I still feel like the optimal solution is to get the current task executor from WorkManager (whether the default or the user-supplied one) and reuse it. There's just no way to do that with only their public API. For everything else I am good with things being sub-optimal if it means sticking with only the public API (e.g. duplicating the SerialExecutor class). However in this one case I do wonder if the tradeoff is worth it. Having an extra thread pool just for our own book keeping is certainly an overhead.

anuchandy: agree, I really wish we have a stable way to  retrieve `WorkManager` executor, which is optimal but so bad that they hide it.

Just double checking, so do you agree with current sub-optimal soln?

1. Internally we use `SerialExecutor `, we don't leak `SerialExecutor ` type through TransferClient public APIs.
2. We allow user to share their `Executor` with us, to which we serially delegate the bookkeeping work.
3. If user don't provide one then we internally create `Executor` and use for book keeping.

bsiegel: #1 yes, I'm fine with us having our own copy of SerialExecutor to avoid having to use non-public API

#2-3, what do you think of this proposal for `TransferClient.build()`:

```java
if (this.serialBackgroundExecutor == null) {
    try {
        // Try to re-use the existing taskExecutor shared by WorkManager and Room
        WorkManagerImpl wmImpl = (WorkManagerImpl)WorkManager.getInstance(context);
        this.serialBackgroundExecutor = new SerialExecutor(wmImpl.getConfiguration().getTaskExecutor());
    } catch (Throwable ignored) {
        // Create our own small ThreadPoolExecutor if we can't
        this.serialBackgroundExecutor = new SerialExecutor(Executors.newFixedThreadPool(2));
    }
}
```

In the above, we're unsafely using the non-public API of WorkManager, but if it fails we catch any type of Throwable and create our own thread pool instead.

anuchandy: cool, I'm fine with doing this best effort to avoid creating our own pool.

So if ever WM remove this accessor then as part of upgrading our libs to the new WM, we can adjust our code.

If customer override WM dependency in their application, such that the version they bring is different from the one we use and if their WM is missing this accessor, then it will be MethodNotFoundException which we catch here (as RuntimeException) and fall back to our own thread pool.

bsiegel: I could also imagine a future change causing ClassCastException or NullPointerException as well but all of those should be caught here. Just documenting for posterity :)

anuchandy:
@bsiegel addressed review comments. Its a force push to the last commit.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -3,93 +3,274 @@
...
+        @SuppressLint("RestrictedApi")
+        public TransferClient build() {
+            if (this.storageBlobClient == null) {
+                throw new IllegalArgumentException("storageBlobClient must be set.");
+            }
+            final Constraints.Builder constraintsBuilder = new Constraints.Builder();
+            constraintsBuilder.setRequiresCharging(this.requiresCharging);
+            constraintsBuilder.setRequiresDeviceIdle(this.requiresDeviceIdle);
+            if (this.networkType == null) {
+                throw new IllegalArgumentException("networkType must be set.");
```
bsiegel: Maybe we should just default this to `CONNECTED` if not specified. That seems like a reasonable default.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -3,93 +3,274 @@
...
+            this.requiresStorageNotLow = requiresStorageNotLow;
+            return this;
+        }
+
+        /**
+         * Set the {@link Executor} used by {@link TransferClient} for all its internal
+         * book keeping, which includes creating DB entries for transfer workers, querying DB
+         * for status, submitting transfer request to {@link WorkManager}.
+         *
+         * TransferClient will enqueue maximum of two command to the serialBackgroundExecutor at any time.
```
bsiegel: Nit:
```suggestion
         * TransferClient will enqueue a maximum of two commands to the taskExecutor at any time.
```

Also for consistency it might be nice to rename the field to e.g. "serialTaskExecutor"

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
anuchandy: hmm, I thought of that a bit worried about it i.e. using a default that can "incur charge/money".

bsiegel: Definitely a good thought. I looked at what WorkManager itself does, and if no Constraints are supplied, the default value is `Constraints.NONE`: https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/work/workmanager/src/main/java/androidx/work/impl/model/WorkSpec.java#97

I'd feel okay doing the same, for consistency. Though I do totally agree that the workmanager is a general framework, while we're using it specifically for network transfers and so the appropriate default might be different for our use case. Wondering if @adrianhall has an opinion here?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
bsiegel: There are a few other comments referring to `serialBackgroundExecutor` that need to be updated as well. I can push a quick fixup or you can go through and take care of it, whichever you prefer.

anuchandy: sure please do, I won't say "no" to help :)

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
anuchandy: Just checked what AWS is doing in this case, aws-lib uses [ANY](https://github.com/aws-amplify/aws-sdk-android/blob/0a4eac920163f80b0e46fe93721a2cc93848f36d/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtilityOptions.java#L184), equivalent that in WM is `CONNECTED`.

anuchandy: Created an issue here: https://github.com/Azure/azure-sdk-for-android/issues/213.

anuchandy:
@bsiegel - Created an issue for the default network type. I'm fine with defaulting to `CONNECTED`.  I'll address that in a later PR based on what we all agree with. 

It seems other than that there are no pending comments to address. Merging this to unblock the next work.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
anuchandy: Based on the response to the issue, we will go with `CONNECTED` as default which is already specified in the code. Marking it resolved.

PR #207, opened 2020-02-27
Improve build

bsiegel:
This PR updates various dependencies, cleans up warnings and errors, and improves the configuration of the build targets. See the individual commit messages for details


build.gradle
```diff
@@ -137,18 +136,15 @@ allprojects {
 
     ext {
         annotationsVersion = "1.1.0"
-        jacksonDatabindVersion = "2.9.9"
-        jacksonDataFormatXmlVersion = "2.9.9"
+        concurrentFuturesVersion = "1.0.0"
         jUnitVersion = "4.12"
-        mockWebServerVersion = "3.13.1"
```
vcolin7: Was this removed this so we use the same version as OkHttp?

bsiegel: Yes, I condensed a few places where we were using different libraries within the same SDK and had different version variables for each library.

samples/sample-app-storage/build.gradle
```diff
@@ -2,7 +2,7 @@ apply plugin: "com.android.application"
 
 android {
     compileSdkVersion 29
-    buildToolsVersion "29.0.2"
+    buildToolsVersion "29.0.3"
```
vcolin7: Is there any reason we are hard-coding this instead of using `buildToolsVersion "$androidBuildToolsVersion"` from root build.gradle file?

samples/sample-app-storage/build.gradle
```diff
@@ -45,15 +46,13 @@ dependencies {
...
     //noinspection GradleDynamicVersion
     implementation "com.microsoft.identity.client:msal:0.3+"
-    implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonDatabindVersion"
-    implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-xml:$jacksonDataFormatXmlVersion"
-    implementation "com.jakewharton.threetenabp:threetenabp:$threeTenAbpVersion"
-    implementation "com.squareup.okhttp3:okhttp:$okHttpVersion"
-    implementation "com.squareup.retrofit2:retrofit:$retrofitVersion"
-    implementation "javax.xml.stream:stax-api:$staxApiVersion" // https://stackoverflow.com/a/47371517/1473510
-    // Unit testing
+    implementation "com.fasterxml.jackson.core:jackson-databind:2.9.9"
```
vcolin7: On the same vein as the buildToolsVersion question, is the reason we hard-coding dependencies here is because we want it to be a complete separate package?

build.gradle
vcolin7: Was this removed this so we use the same version as OkHttp?

samples/sample-app-storage/build.gradle
vcolin7: Is there any reason we are hard-coding this instead of using buildToolsVersion "$androidBuildToolsVersion" like in our root build.gradle file?

samples/sample-app-storage/build.gradle
vcolin7: On the same vein as the buildToolsVersion question, is the reason we hard-coding dependencies here is because we want it to be a complete separate package?

build.gradle
bsiegel: Yes, I condensed a few places where we were using different libraries within the same SDK and had different version variables for each library.

samples/sample-app-storage/build.gradle
bsiegel: I went back and forth on this, but since this is a sample, I assume we would want a customer to be able to download the project and open it up and have it just work. It's a maintenance burden to keep these dependency versions up to date, but I think it would be more confusing to the customer if they were just all referencing variables defined elsewhere.

Actually given that, we should probably use a local copy of the lint.xml file too. I'll fix that shortly.

samples/sample-app-storage/build.gradle
bsiegel: Addressed above

sdk/storage/ci.yml
```diff
@@ -37,8 +37,3 @@ stages:
   - template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml
     parameters:
       ServiceDirectory: storage
```
mitchdenny: Removing the artifacts stops the release pipelines being able to publish the package.

Issue #210, opened 2020-02-29
Issues with CurlLoggingInterceptor


vcolin7:
Found a number of issues with the CurlLoggingInterceptor:

- A placeholder message will be logged if the request body is empty, not inline, not identity (`gzip`, `compressed`, etc.) or binary, instead of logging the actual contents of the body.
- Double quotation marks surround the URL in the cURL command, which is not necessary.
- Double quotation marks in header names are escaped instead of escaping them in the header values.
- Backslashes are not escaped in header values.
- The `--compressed` flag is only added when Content-Encoding of the body is `gzip`, instead of adding it when the encoding is not `identity`.
- Double quotation marks are escaped in the body instead of single quotes.

PR #211, opened 2020-03-01
Fix interceptor bugs uncovered by tests

### LINKED ISSUE #209: Issues with LoggingInterceptor
Found a number of issues with the LoggingInterceptor:

- The body of a request or response is not logged if the Content-Type is 'identity' or Content-Encoding is 'inline', when actually we should not log it when the value is different from those, like 'gzip', 'attached' or 'compressed'.
- The body of a request or response is logged when the Content-Length header is empty or not present, instead of using a placeholder message.
- The body of a request or response body is not logged if its size is smaller than the maximum value, when it should be the opposite (size > MAX_VALUE).
- A malformed query string is generated when redacting sensitive data (it's missing the starting '?' character).
- An UnsupportedOperationException is thrown when trying to add elements to the allowed headers and query parameters sets. This is due to using an immutable empty set instead of a mutable one.
- Internal failures (exceptions) are not caught when logging a request, only when logging the response.

vcolin7:

Fixes #209 & #210.

- Corrected the behavior of the LoggingInterceptor to adhere to the design guidelines.
- Corrected the behavior of the CurlLoggingInterceptor to adhere to the design guidelines.
- Subsequently made some changes to the LogUtils class considering the aforementioned.

More details can be found in each commit message of this PR.


sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogUtils.java
```diff
@@ -40,33 +30,45 @@ static long getContentLength(Headers headers) {
      * @return "Log body" if the body should be logged in its entirety, otherwise a message indicating why the body
      * was not logged is returned.
      */
-    static String evaluateBody(Headers headers) {
+    @SuppressWarnings("ConstantConditions")
```
bsiegel: What was causing this warning? It doesn't seem like it should be triggering.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LoggingInterceptor.java
```diff
@@ -96,24 +103,42 @@ public Response intercept(@NonNull Chain chain) throws IOException {
...
-        String bodyEvaluation = LogUtils.evaluateBody(request.headers());
         RequestBody requestBody = request.body();
+        boolean contentTypeLogged = false;
+        boolean contentLengthLogged = false;
+
+        // According to the OkHttp documentation, the request body headers are only present when this is installed as a
+        // network interceptor.
+        // https://github.com/square/okhttp/blob/b189a382bccc1b9a01d4672210b69680d73b4306/okhttp-logging-interceptor/src/main/java/okhttp3/logging/HttpLoggingInterceptor.java#L173
+        if (requestBody != null) {
+            if (requestBody.contentType() != null  && allowedHeaderNames.contains(CONTENT_TYPE_HEADER)) {
```
bsiegel: Why do these headers get special treatment?

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogUtils.java
```diff
@@ -40,33 +30,45 @@ static long getContentLength(Headers headers) {
      * @return "Log body" if the body should be logged in its entirety, otherwise a message indicating why the body
```
bsiegel: Definitely need to get rid of this magic string. Even if we made this into a constant, it's still a little suspicious. Why not name this method something like `getBodySummary()` and then if it returns `null` log the body text, otherwise log the summary?

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LogUtils.java
bsiegel: Going to check out this PR and see what the heck is going on there because I can't see any reason why that would generate this warning.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/LoggingInterceptor.java
bsiegel: If the Content-Type or Content-Length are not present in the headers, we don't need to "synthesize" them for logging purposes.

The only reason I could see needing to do that is if OkHTTP itself adds those headers, but it hasn't done so yet (by the time our logging interceptor is called). Which is pretty annoying if that's the case, since interceptors should be able to assume the headers list is finalized by the time they receive the request or response. If that's what's going on, we can keep this code but I think we could probably refactor it a bit to make it more clear. I'll poke at this while I have the PR checked out as mentioned above.

bsiegel:
Pushed a commit with some improvements, it couldn't get quite as clean as I wanted but I think it's reasonably good. Not sure if it still passes the tests though, I will try to reincorporate the test PR into this one locally and run them.

PR #214, opened 2020-03-10
Unit testing core util

### LINKED ISSUE #189: Increase code coverage of unit tests for Preview 1

vcolin7:

Partially addresses #189.

- Increased test coverage for package `com.azure.android.core.util`.

This considers changes made in #215.

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java
@@ -13,6 +13,13 @@ public final class Base64Util {
         // Empty constructor to prevent instantiation of this class.
     }
 
+    /**
+     * We cannot call Android framework's classes like {@link Base64} from unit tests and we cannot use Java's
+     * Base64 class because that's only available at a later Android version. To get around this and not use
+     * something like PowerMock to mock static methods like {@code Base64.encode()}, the {@link Base64Wrapper}
+     * interface was created. Normally this value will always be set to {@link Base64Android} and for unit tests it will
+     * be set to a stub that implements the aforementioned interface.
+     */
     private static Base64Wrapper base64Wrapper = new Base64Android();
 
     static void setBase64Wrapper(Base64Wrapper newWrapper) {
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java
@@ -3,7 +3,6 @@
 
 package com.azure.android.core.util;
 
-
 import android.util.Base64;
 
 /**
@@ -14,6 +13,12 @@ public final class Base64Util {
         // Empty constructor to prevent instantiation of this class.
     }
 
+    private static Base64Wrapper base64Wrapper = new Base64Android();
+
+    static void setBase64Wrapper(Base64Wrapper newWrapper) {
+        base64Wrapper = newWrapper;
+    }
+
     /**
      * Encodes a byte array to Base64.
      *
@@ -21,7 +26,7 @@ public final class Base64Util {
      * @return The Base64 encoded bytes.
      */
     public static byte[] encode(byte[] src) {
-        return src == null ? null : Base64.encode(src, Base64.DEFAULT);
+        return src == null ? null : base64Wrapper.encode(src, Base64.DEFAULT);
     }
 
     /**
@@ -33,7 +38,7 @@ public final class Base64Util {
     public static byte[] encodeUrlWithoutPadding(byte[] src) {
         int flags = Base64.URL_SAFE | Base64.NO_PADDING;
 
-        return src == null ? null : Base64.encode(src, flags);
+        return src == null ? null : base64Wrapper.encode(src, flags);
     }
 
     /**
@@ -43,7 +48,7 @@ public final class Base64Util {
      * @return The Base64 encoded bytes.
      */
     public static String encodeToString(byte[] src) {
-        return src == null ? null : Base64.encodeToString(src, Base64.DEFAULT);
+        return src == null ? null : base64Wrapper.encodeToString(src, Base64.DEFAULT);
     }
 
     /**
@@ -53,7 +58,7 @@ public final class Base64Util {
      * @return The decoded byte array.
      */
     public static byte[] decode(byte[] encoded) {
-        return encoded == null ? null : Base64.decode(encoded, Base64.DEFAULT);
+        return encoded == null ? null : base64Wrapper.decode(encoded, Base64.DEFAULT);
     }
 
     /**
@@ -63,7 +68,7 @@ public final class Base64Util {
      * @return The decoded byte array.
      */
     public static byte[] decodeUrl(byte[] src) {
-        return src == null ? null : Base64.decode(src, Base64.URL_SAFE);
+        return src == null ? null : base64Wrapper.decode(src, Base64.URL_SAFE);
     }
 
     /**
@@ -73,7 +78,7 @@ public final class Base64Util {
      * @return The decoded byte array.
      */
     public static byte[] decodeUrl(String src) {
-        return src == null ? null : Base64.decode(src, Base64.URL_SAFE);
+        return src == null ? null : base64Wrapper.decode(src, Base64.URL_SAFE);
     }
 
     /**
@@ -83,6 +88,35 @@ public final class Base64Util {
      * @return The decoded byte array.
      */
     public static byte[] decodeString(String encoded) {
-        return encoded == null ? null : Base64.decode(encoded, Base64.DEFAULT);
+        return encoded == null ? null : base64Wrapper.decode(encoded, Base64.DEFAULT);
+    }
+
+    interface Base64Wrapper {
+        byte[] encode(byte[] input, int flags);
+        String encodeToString(byte[] input, int flags);
+        byte[] decode(byte[] input, int flags);
+        byte[] decode(String input, int flags);
+    }
+
+    static class Base64Android implements Base64Wrapper {
+        @Override
+        public byte[] encode(byte[] input, int flags) {
+            return Base64.encode(input, flags);
+        }
+
+        @Override
+        public String encodeToString(byte[] input, int flags) {
+            return Base64.encodeToString(input, flags);
+        }
+
+        @Override
+        public byte[] decode(byte[] input, int flags) {
+            return Base64.decode(input, flags);
+        }
+
+        @Override
+        public byte[] decode(String input, int flags) {
+            return Base64.decode(input, flags);
+        }
     }
 }
--- a/sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
@@ -11,6 +11,8 @@ import org.junit.runner.RunWith;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
 
+import java.util.Arrays;
+
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -186,6 +188,30 @@ public class Base64UrlTest {
         assertArrayEquals(ENCODED_BYTES, base64Url.encodedBytes());
         assertArrayEquals(DECODED_BYTES, base64Url.decodedBytes());
         assertEquals("AAECAwQFBgcICQ", base64Url.toString());
+    }
+
+    @Test
+    public void toString_withSimpleString() {
+        final Base64Url base64Url = new Base64Url("AAECAwQFBgcICQ");
+
+        assertEquals("AAECAwQFBgcICQ", base64Url.toString());
+    }
+
+    @Test
+    public void toString_withNullString() {
+        final Base64Url base64Url = new Base64Url((String) null);
+
+        assertEquals("", base64Url.toString());
+    }
+
+    @Test
+    public void hashCode_equalToEncodedBytesHashCode() {
+        spy(Base64.class);
+        when(Base64.encode(notNull(), anyInt())).thenReturn(ENCODED_BYTES);
+
+        final Base64Url base64Url = Base64Url.encode(DECODED_BYTES);
+
+        assertEquals(Arrays.hashCode(ENCODED_BYTES), base64Url.hashCode());
     }
 
     @SuppressWarnings("SimplifiableJUnitAssertion")
--- a/sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
+++ b/sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
@@ -3,38 +3,24 @@
 
 package com.azure.android.core.util;
 
-import android.util.Base64;
-
-import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.powermock.core.classloader.annotations.PrepareForTest;
-import org.powermock.modules.junit4.PowerMockRunner;
 
 import java.util.Arrays;
 
+import static com.azure.android.core.util.Base64Stub.DECODED_BYTES;
+import static com.azure.android.core.util.Base64Stub.ENCODED_BYTES;
+import static com.azure.android.core.util.Base64Stub.ENCODED_STRING;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyInt;
-import static org.mockito.ArgumentMatchers.notNull;
-import static org.mockito.Mockito.when;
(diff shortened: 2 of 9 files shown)


bsiegel:
If possible I'd like to find a way to keep mocking frameworks out of our library. Generally unless it's something coupled to the Android OS we should be able to redesign the internals of whatever it is we're testing to avoid them.

vcolin7:
@bsiegel I'll try to spend some time thinking if we can get around using PowerMock or Mockito for the Base64 classes.

vcolin7:
@bsiegel Also, I'm curious as to why we want to avoid mocking libraries in our tests

sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java
```diff
@@ -14,14 +13,20 @@ private Base64Util() {
         // Empty constructor to prevent instantiation of this class.
     }
 
+    private static Base64Wrapper base64Wrapper = new Base64Android();
```
alzimmermsft: This is for testing purposes only right? If not should this be atomic, or locked when modified, as this isn't completely thread-safe right now.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java
```diff
@@ -83,6 +88,35 @@ public static String encodeToString(byte[] src) {
      * @return The decoded byte array.
      */
     public static byte[] decodeString(String encoded) {
-        return encoded == null ? null : Base64.decode(encoded, Base64.DEFAULT);
+        return encoded == null ? null : base64Wrapper.decode(encoded, Base64.DEFAULT);
+    }
+
+    interface Base64Wrapper {
```
alzimmermsft: Is this planned to be used in the public API? If so, instead of maintaining context in this static helper class should a new overload to each API be added which accepts this interface? That way this class remains thread safe and stateless as expected with a static helper class.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
```diff
@@ -0,0 +1,201 @@
...
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class Base64UrlTest {
+    @BeforeClass
+    public static void setUp() {
+        Base64Util.setBase64Wrapper(new Base64Stub());
```
alzimmermsft: Is a wrapper being set as Base64 isn't consistent between Android versions or between devices?

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
```diff
@@ -0,0 +1,201 @@
...
+        assertEquals(Arrays.hashCode(ENCODED_BYTES), base64Url.hashCode());
+    }
+
+    @SuppressWarnings("SimplifiableJUnitAssertion")
+    @Test
+    public void equalTo_identicalObject() {
+        final Base64Url base64Url = new Base64Url((byte[]) null);
+        final Base64Url otherBase64Url = new Base64Url((byte[]) null);
+
+        assertTrue(base64Url.equals(otherBase64Url));
```
alzimmermsft: Any reason that `assertEquals` can't be used here? It would allow for the suppression to be removed.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    @SuppressWarnings("SimplifiableJUnitAssertion")
```
alzimmermsft: Same question

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
```diff
@@ -0,0 +1,201 @@
...
+
+        assertFalse(base64Url.equals(otherBase64Url));
+    }
+
+    @SuppressWarnings({"SimplifiableJUnitAssertion", "ConstantConditions"})
+    @Test
+    public void notEqualTo_nullObject() {
+        final Base64Url base64Url = new Base64Url((byte[]) null);
+
+        assertFalse(base64Url.equals(null));
```
alzimmermsft: Does `Base64Url`'s `equals` overload compare against the `String` representation? If so could this be `assertNull(base64Url.toString())`?

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    @SuppressWarnings({"SimplifiableJUnitAssertion", "EqualsBetweenInconvertibleTypes"})
+    @Test
+    public void notEqualTo_objectFromDifferentClass() {
+        final Base64Url base64Url = new Base64Url((byte[]) null);
+
+        assertFalse(base64Url.equals(""));
```
alzimmermsft: Same comment as `notEqualTo_nullObject`.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UtilTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test
+    public void decodeUrl_fromNullByteArray() {
+        assertNull(Base64Util.decodeUrl((String) null));
+    }
+
+    @Test
+    public void decodeUrl_fromNullString() {
+        assertNull(Base64Util.decodeUrl((String) null));
+    }
```
alzimmermsft: I think some tests should be added that use non-null, right now this will get good code coverage but is missing real world scenarios.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/CoreUtilTest.java
```diff
@@ -0,0 +1,112 @@
...
+    public void isNullOrEmpty_returnsFalse_withNonEmptyString() {
+        assertFalse(isNullOrEmpty("Test"));
+    }
+
+    @Test
+    public void replaceWithSingleCharacters_inString() {
+        Map<Character, CharSequence> pairs = new HashMap<>();
+        pairs.put('a', "A");
+        pairs.put(' ', "_");
+        String testString = "Ohana means family.";
```
alzimmermsft: ❤️ 😄

sdk/core/azure-core/src/test/java/com/azure/android/core/util/DateTimeRfc1123Test.java
```diff
@@ -0,0 +1,91 @@
...
+
+        assertEquals(TEST_DATE, dateTimeRfc1123.toString());
+    }
+
+    @Test
+    public void getHashCode() {
+        DateTimeRfc1123 dateTimeRfc1123 = new DateTimeRfc1123(
+            OffsetDateTime.of(LocalDateTime.parse(TEST_DATE, RFC1123_DATE_TIME_FORMATTER), ZoneOffset.UTC));
+
+        assertEquals(1468106660, dateTimeRfc1123.hashCode());
```
alzimmermsft: Unless we have overridden the hash code method I wouldn't eagerly test this as it is dependent on the underlying JVM and has the potential for collisions.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/DateTimeRfc1123Test.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @SuppressWarnings("SimplifiableJUnitAssertion")
+    @Test
+    public void equalTo_identicalObject() {
+        final DateTimeRfc1123 dateTimeRfc1123 = new DateTimeRfc1123(
+            OffsetDateTime.of(LocalDateTime.parse(TEST_DATE, RFC1123_DATE_TIME_FORMATTER), ZoneOffset.UTC));
+        final DateTimeRfc1123 otherDateTimeRfc1123 = new DateTimeRfc1123(
+            OffsetDateTime.of(LocalDateTime.parse(TEST_DATE, RFC1123_DATE_TIME_FORMATTER), ZoneOffset.UTC));
+
+        assertTrue(dateTimeRfc1123.equals(otherDateTimeRfc1123));
```
alzimmermsft: `assertEquals`?

sdk/core/azure-core/src/test/java/com/azure/android/core/util/ExpandableStringEnumTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Test
+    public void convertToString() {
+        assertEquals("Test", MyExpandableStringEnum.fromString("Test").toString());
+    }
+
+    @Test
+    public void getHashCode() {
+        assertEquals(-2043209345, MyExpandableStringEnum.fromString("Test").hashCode());
+    }
```
alzimmermsft: Same comment as the other `hashCode` test.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/ExpandableStringEnumTest.java
```diff
@@ -0,0 +1,124 @@
...
+        MyExpandableStringEnum otherMyExpandableStringEnum = MyExpandableStringEnum.fromString("Another");
+        Collection instances = MyExpandableStringEnum.values();
+
+        assertFalse(instances.isEmpty());
+        assertEquals(2, instances.size());
+
+        Iterator iterator = instances.iterator();
+
+        assertSame(myExpandableStringEnum, iterator.next());
+        assertSame(otherMyExpandableStringEnum, iterator.next());
```
alzimmermsft: How strong is the guarantee that the backing `Map` for `ExandableStringEnum` will return a collection where the elements are ordered by how they are added?

sdk/core/azure-core/src/test/java/com/azure/android/core/util/ExpandableStringEnumTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+        assertFalse(myExpandableStringEnum.equals(otherMyExpandableStringEnum));
+    }
+
+    @SuppressWarnings({"SimplifiableJUnitAssertion", "ConstantConditions"})
+    @Test
+    public void notEqualTo_nullObject() {
+        MyExpandableStringEnum myExpandableStringEnum = MyExpandableStringEnum.fromString("Test");
+
+        assertFalse(myExpandableStringEnum.equals(null));
```
alzimmermsft: `assertNotNull`?

sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java
vcolin7: It is only for testing purposes. I probably should add a clarifying comment though.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java
vcolin7: This is not planned to be a part of the public API. The reason I wrote this interface was so we wouldn't have to mock the static methods in Android's Base64 class. This is due to a couple of reasons:

- Unfortunately, we cannot use Java's version of this class since it only works with a higher Android version than our supported minimum.
- We cannot access the Android framework's classes in unit tests, which are only accessible when running on the actual operating system on an emulator or physical device. Unit tests run on the local JVM.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
vcolin7: There is no reason not to use `assertEquals`. I must have missed it. Thanks for pointing it out!

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
vcolin7: See answer above.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/DateTimeRfc1123Test.java
vcolin7: Agreed.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/DateTimeRfc1123Test.java
vcolin7: We did override it in our `DateTimeRfc1123` class, however, it just uses the `hashCode` method in `OffsetDateTime`. I think I will remove this test for the time being.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/ExpandableStringEnumTest.java
vcolin7: See comment above for the same question.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/ExpandableStringEnumTest.java
vcolin7: Agreed.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/ExpandableStringEnumTest.java
vcolin7: There is no guarantee since this is not a LinkedHashSet, so I think the best course of action would be to cast the collectionreturned by `values()` to a HashSet and see if the values we are looking for are contained there.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/DateTimeRfc1123Test.java
vcolin7: Actually, after digging around on the internet, I think I will take the approach showed [here](https://stackoverflow.com/questions/4449728/how-can-i-do-unit-test-for-hashcode).

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
vcolin7: The reason to set this here is that we cannot call Android framework's classes like Base64 from unit tests and we cannot use Java's Base64 class because that's only available at a later Android version. To get around this and not use something like PowerMock to mock static methods like Base64.encode(), I created the [Base64Wrapper](https://github.com/Azure/azure-sdk-for-android/blob/67845f0155a5700833dae754fbd7c710ca8aaddb/sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java#L94) interface. Normally it will always be set to [Base64Android](https://github.com/Azure/azure-sdk-for-android/blob/67845f0155a5700833dae754fbd7c710ca8aaddb/sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java#L101) and for unit tests it will be set to [Base64Stub](https://github.com/Azure/azure-sdk-for-android/blob/67845f0155a5700833dae754fbd7c710ca8aaddb/sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64Stub.java).

sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java
vcolin7: Would it be better to move the interface and class below to their own files?

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
vcolin7: You are right, this can be made to use `assertNull`.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UrlTest.java
vcolin7: Here we can just use `assertNotEquals`.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/Base64UtilTest.java
vcolin7: Sounds good, will add some tests that uses `assertNonNull`.

vcolin7: Based on the test names, it actually seems I was planning to have the non-null tests but somehow forgot about it.

vcolin7: I will correct and add tests that check for the positive scenarios when we have a non-null source and non-empty string.

vcolin7:
> 
> 
> Are any of these tests portable to `azure-sdk-for-java` and cover gaps in testing there? If so would you mind opening either an issue or a PR against the repo and either copy or reference this PR.

I'm pretty sure some of these should be portable to the Java repo. I'll create a PR later during the week.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/Base64Util.java
alzimmermsft: Leaving them in this file is fine, I just didn't have a complete understanding of this when I wrote the comment.

PR #224, opened 2020-04-13
Error notification, TransferInfo LiveData sharing and pause, resume support

anuchandy:
#### Features:

1. Adding support to promote errors from Background Worker and SerialExecutor to TransferInfo LiveData
    - From Background Worker https://github.com/Azure/azure-sdk-for-android/commit/e3f433ea459b818fdb53952102dbebed53e033ea
    - From SerialExecutor https://github.com/Azure/azure-sdk-for-android/pull/224/commits/7f18255f1119be4439ef13bd552a11bc07c23087
2. Adding support for weak-reference based cache to enable sharing of TransferInfo LiveData per transfer - commit https://github.com/Azure/azure-sdk-for-android/commit/68ee38e04e403bc8a2fb3984ecc39e19c8c2d2d4
3. Adding support for pause, resume infra and enabling it for upload - commit https://github.com/Azure/azure-sdk-for-android/commit/189d1be3d051aa840ae3d5e9b228b7fb59050d6f

#### Refactor:

1. Renaming TransferIdMappedToTransferInfo to TransferIdInfoLiveData - commit https://github.com/Azure/azure-sdk-for-android/commit/d696b38f95e3b44981986e36910786ba97b6bc5f

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
@@ -77,7 +77,7 @@ public class TransferClient {
      */
     public LiveData<TransferInfo> upload(String containerName, String blobName, File file) {
         // UI_Thread
-        final MutableLiveData<TransferIdOrError> idOrErrorChannel = new MutableLiveData<>();
+        final MutableLiveData<TransferIdOrError> idOrErrorLiveData = new MutableLiveData<>();
         this.serialTaskExecutor.execute(() -> {
             // BG_Thread
             try {
@@ -104,19 +104,19 @@ public class TransferClient {
                         ExistingWorkPolicy.KEEP,
                         uploadWorkRequest)
                     .enqueue();
-                idOrErrorChannel.postValue(TransferIdOrError.id(TransferIdOrError.Operation.UPLOAD_DOWNLOAD, transferId));
+                idOrErrorLiveData.postValue(TransferIdOrError.id(TransferIdOrError.Operation.UPLOAD_DOWNLOAD, transferId));
             } catch (Exception e) {
-                idOrErrorChannel.postValue(TransferIdOrError.error(TransferIdOrError.Operation.UPLOAD_DOWNLOAD, e));
+                idOrErrorLiveData.postValue(TransferIdOrError.error(TransferIdOrError.Operation.UPLOAD_DOWNLOAD, e));
             }
         });
         // UI_Thread
-        return toCachedTransferInfoLiveData(idOrErrorChannel, false);
+        return toCachedTransferInfoLiveData(idOrErrorLiveData, false);
     }
 
     /**
      * Pause a transfer identified by the given transfer id. The pause operation
-     * is a best-effort and it is possible that a transfer that is already executing
-     * may continue to transfer.
+     * is a best-effort, and a transfer that is already executing may continue to
+     * transfer.
      *
      * Upon successful scheduling of the pause, any observer observing on
      * {@link LiveData<TransferInfo>} for this transfer receives a {@link TransferInfo}
@@ -152,10 +152,10 @@ public class TransferClient {
     }
 
     /**
-     * Resume a transfer that was paused.
+     * Resume a paused transfer.
      *
-     * @param @param transferId the transfer id identifies the transfer to resume.
-     * @return LiveData that streams {@link TransferInfo} describing current state of the transfer
+     * @param transferId the transfer id identifies the transfer to resume.
+     * @return LiveData that streams {@link TransferInfo} describing the current state of the transfer
      */
     public LiveData<TransferInfo> resume(long transferId) {
         // UI_Thread
@@ -180,12 +180,10 @@ public class TransferClient {
                     } else {
                         throw new RuntimeException("Download::resume() NotImplemented");
                     }
-                    //
-                    // Resume will resubmit the work to WorkManager with policy as KEEP.
-                    // which means if the work is already running then this resume() call
-                    // is NO-OP. We return the LiveData to the caller that can be observed
-                    // for TransferInfo events of the already running work.
-                    //
+                    // resume() will resubmit the work to WorkManager with the policy as KEEP.
+                    // With this policy, if the work is already running, then this resume() call is NO-OP,
+                    // we return the LiveData to the caller that streams the TransferInfo events of
+                    // the already running work.
                     WorkManager.getInstance(context)
                         .beginUniqueWork(toTransferUniqueWorkName(transferId),
                             ExistingWorkPolicy.KEEP,
@@ -215,18 +213,18 @@ public class TransferClient {
      * Subscribe to a TransferIdOrError LiveData and transform that to TransferInfo LiveData.
      *
      * This method caches or uses cached {@link LiveData<TransferInfo>} to stream TransferInfo.
-     * If provided TransferIdOrError LiveData emit an error then cache won't be used.
+     * If provided TransferIdOrError LiveData emits an error, then cache won't be used.
      *
-     * @param idOrErrorChannel the TransferIdOrError LiveData
+     * @param idOrErrorLiveData the TransferIdOrError LiveData
      * @param isResume true if the transfer id emitted by the TransferIdOrError LiveData identifies
      *                 a transfer to be resumed, false for a new upload or download transfer.
      * @return the TransferInfo LiveData
      */
     @MainThread
-    private LiveData<TransferInfo> toCachedTransferInfoLiveData(LiveData<TransferIdOrError> idOrErrorChannel,
+    private LiveData<TransferInfo> toCachedTransferInfoLiveData(LiveData<TransferIdOrError> idOrErrorLiveData,
                                                                 boolean isResume) {
         // UI_Thread
-        return Transformations.switchMap(idOrErrorChannel, idOrError -> {
+        return Transformations.switchMap(idOrErrorLiveData, idOrError -> {
             if (idOrError.isError()) {
                 final TransferIdInfoLiveData.Result result = TransferIdInfoLiveData.create(context);
                 final TransferIdInfoLiveData.LiveDataPair pair = result.getLiveDataPair();
@@ -234,17 +232,20 @@ public class TransferClient {
                 return pair.getTransferInfoLiveData();
             } else {
                 if (isResume) {
-                    // If application process already has a cached LiveData pair for the same transfer
-                    // then use it, otherwise create, cache and use.
+                    // If the application process already has a cached LiveData pair for the same transfer,
+                    // then use it, otherwise create, cache, and use.
                     final TransferIdInfoLiveData.LiveDataPair pair
                         = TRANSFER_ID_INFO_CACHE.getOrCreate(idOrError.getId(), context);
                     pair.getTransferIdOrErrorLiveData().setValue(idOrError);
                     return pair.getTransferInfoLiveData();
                 } else {
-                    // Create a transfer LiveData pair cache entry and use it.
-                    // Any future resume operation on the same transfer will use this pair as long as
-                    // both operations (upload_download and resume) happened in the same application
-                    // process and the cache entry is not GC-ed.
+                    // For a new upload or download transfer, create a transfer LiveData pair
+                    // (TransferIdOrError, TransferInfo) cache entry and use them.
+                    // Any future resume operation on the same transfer will use this pair
+                    // as long as:
+                    //     1. both upload or download transfer, and the corresponding resume happens
+                    //        in the same application process
+                    //     2. and the cache entry is not GC-ed.
                     final TransferIdInfoLiveData.LiveDataPair pair
                         = TRANSFER_ID_INFO_CACHE.create(idOrError.getId(), context);
                     pair.getTransferIdOrErrorLiveData().setValue(idOrError);
@@ -257,7 +258,7 @@ public class TransferClient {
     /**
      * Do pre-validations to see a transfer can be resumed.
      *
-     * @param transferId the transfer id to check for resume eligibility
+     * @param transferId identifies the transfer to check for resume eligibility
      * @param idOrErrorChannel the LiveData to post the error if the transfer cannot be resumed
      * @return result of check
      */
@@ -297,7 +298,7 @@ public class TransferClient {
     /**
      * Do pre-validations to see a transfer can be paused.
      *
-     * @param transferId the transfer id to check for pause eligibility
+     * @param transferId identifies the transfer to check for pause eligibility
      * @return result of check
      */
     private PauseCheck checkPauseable(long transferId) {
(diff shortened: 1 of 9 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
```diff
@@ -0,0 +1,477 @@
...
+    // flag to track whether current event about to emit from transferInfoLiveData is the first event.
+    private boolean isFirstEvent = true;
+    // flag to track whether the last emitted event from transferInfoLiveData was a pause event.
+    private boolean wasPaused;
+
+    private TransferIdInfoLiveData() {}
+
+    @MainThread
+    static TransferIdInfoLiveData.Result create(Context context) {
+        TransferIdInfoLiveData transferIdInfoLiveData = new TransferIdInfoLiveData();
```
vcolin7, line 63: This may be a dumb question but why are we using a static method to create this instead of making `init()` not private and using it?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
```diff
@@ -0,0 +1,477 @@
...
+        /**
+         * Check whether the user paused the transfer in the current session.
+         * Note: this method reset the flag once called.
+         *
+         * @return true if user paused the transfer, false otherwise.
+         */
+        @MainThread
+        boolean isUserPaused() {
+            boolean b = this.userPaused;
+            this.userPaused = false;
```
vcolin7, line 473: Why is this reset when checking its value?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveDataCache.java
```diff
@@ -0,0 +1,220 @@
...
+ * same application process. This enables all Observers of a transfer [e.g. Observers of
+ * upload(tid: 1), Observers of resume(tid: 1)] to listen to same TransferInfo LiveData,
+ * irrespective of the TransferClient they used to get the TransferInfo LiveData.
+ *
+ * The cache implementation assumes that the TransferIdOrError LiveData in the value
+ * do not strongly refer to the TransferInfo LiveData in the same value either
+ * directly or indirectly, since that will prevent the TransferIdOrError LiveData
+ * from being collected.
+ */
+final public class TransferIdInfoLiveDataCache {
```
vcolin7, line 47: What is was the reasoning behind making this class public? It seems it is only used within this package. Is the decision related to how the GC works?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -73,39 +76,133 @@ private TransferClient(Context context,
...
+
+    /**
+     * Resume a transfer that was paused.
+     *
+     * @param @param transferId the transfer id identifies the transfer to resume.
+     * @return LiveData that streams {@link TransferInfo} describing current state of the transfer
+     */
+    public LiveData<TransferInfo> resume(long transferId) {
+        // UI_Thread
+        final MutableLiveData<TransferIdOrError> idOrErrorChannel = new MutableLiveData<>();
```
vcolin7, line 162: Is it a convention to name MutableLiveData objects channels?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
```diff
@@ -0,0 +1,477 @@
...
+                // An error from TransferClient then emit null WorkInfo. The downstream should check error before
+                // start processing the WorkInfo.
+                return null;
+            } else {
+                final long transferId = this.inputTransferIdOrError.getId();
+                if (workInfoList == null || workInfoList.isEmpty()) {
+                    Log.e(TAG, "Received null or Empty WorkInfo list for the transfer '" + transferId + "' from WorkManager." );
+                    return null;
+                }
+                if (workInfoList.size() > 1) {
```
vcolin7, line 274: When would we get more than one WorkInfo? When the operation has been paused and resumed multiple times for example? If so, why do we only return the first one? Does that mean they all contain the same info?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveDataCache.java
anuchandy: good catch! this must be a package-private, being public is a type.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
anuchandy: there is no convention like that, just my own convention, memories of Go :) We can remove that suffix.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
anuchandy: WorkManager supports chaining of workers and assigning a name to the chain, in such cases this WorkManager API returns a List of WorkInfo. In our case chaining is not used and each transfer/work has a unique name.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
anuchandy: No specific reasons just to view it as one time consumable value.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/Util.java
```diff
@@ -0,0 +1,66 @@
...
+import com.azure.android.core.internal.util.serializer.SerializerAdapter;
+import com.azure.android.core.internal.util.serializer.SerializerFormat;
+import com.azure.android.core.util.CoreUtil;
+import com.azure.android.storage.blob.models.BlobStorageException;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement;
+
+import java.util.Objects;
+
+final class Util {
```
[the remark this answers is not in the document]
anuchandy: My thought was to use `Util` as a place holder to add any more utility methods that we want to stay local to the package. If we come across more exception parsing scenarios then I was thinking of creating a type like `ExceptionUtil` specific for exception parsing. But I don't see any immediate use case to parse exceptions in any other format.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
anuchandy: Mainly for readability, the name `create` indicates it is factory method hence is static but `init` is initializing an already created object.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
anuchandy: and Nothing breaks if we remove it :)

anuchandy: now I remember why it is a one-time consumable value because if there is a `cancellation` after a `pause`, we don't want to treat that as a `pause`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdOrError.java
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * The type used by the methods in {@link TransferClient} (e.g. upload, download, resume) to
+ * communicate data from background {@link SerialExecutor} to {@link TransferIdMappedToTransferInfo}.
+ *
+ * The data includes the operation type, the actual transfer id of a transfer or an error indicating
+ * the failure detected by TransferClient. Note that this type is used only to channel the error that
+ * TransferClient methods encountered or could detect, this type is not used to channel the real time
+ * error happened in the background worker.
+ */
+public final class TransferIdOrError {
```
bsiegel: I feel like maybe we should just make a generic `Result<TSuccess, TErr>` type we can use instead of this. I discovered that this is somehow not a thing that Java includes itself.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
```diff
@@ -35,6 +35,9 @@
     private final MutableLiveData<TransferIdOrError> transferIdOrErrorLiveData = new MutableLiveData<>();
     // the output TransferInfo LiveData.
     private final MediatorLiveData<TransferInfo> transferInfoLiveData = new MediatorLiveData<>();
+    // the the object that TransferClient methods update any state that it want TransferInfo LiveData source
```
bsiegel, line 48: I don't understand this comment, can you maybe reword? It's not clear to me at this point what the TransferFlags are for. Maybe this encapsulates the different states a transfer can be in?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveDataCache.java
```diff
@@ -0,0 +1,218 @@
...
+        }
+    }
+
+    /**
+     * Expunges stale entries (containing already GCed {@code TransferInfo} LiveData) from the map.
+     */
+    @SuppressWarnings("unchecked")
+    private void expunge() {
+        int cnt = 0;
+        for (Reference<?> staledReference; (staledReference = this.queue.poll()) != null;) {
```
bsiegel, line 140: Is this not just the same as a while loop? I'm fine with it but seeing this for-loop syntax is more confusing to me.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -101,16 +104,105 @@ private TransferClient(Context context,
...
-        return liveDataPair.getTransferInfoLiveData();
+        // UI_Thread
+        return toCachedTransferInfoLiveData(idOrErrorChannel, false);
     }
 
     /**
-     * Get the name for a unique transfer work.
+     * Pause a transfer identified by the given transfer id. The pause operation
+     * is a best-effort and it is possible that a transfer that is already executing
+     * may continue to transfer.
```
bsiegel: What causes the behavior described?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
bsiegel: I like the `channel` nomenclature myself. It's easier for me to think of `LiveDatas` as just channels.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
```diff
@@ -84,6 +101,15 @@ private TransferIdInfoLiveData() {}
                 this.setWasPaused(false);
                 this.transferInfoLiveData.setValue(TransferInfo.createStarted(transferId));
                 if (currentWorkInfoState == WorkInfo.State.ENQUEUED) {
+                    // the worker state can be WorkInfo.State.ENQUEUED in two cases -
```
bsiegel, line 105: ❤️

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
```diff
@@ -241,20 +285,44 @@ private void setLastWorkInfoState(WorkInfo.State state) {
...
+                        this.setLastWorkInfoState(WorkInfo.State.SUCCEEDED);
+                        return;
+                    }
+                }
+                this.transferInfoLiveData.setValue(TransferInfo.createFailed(this.inputTransferIdOrError.getId(),
+                    this.inputTransferIdOrError.getErrorMessage()));
+                this.setLastWorkInfoState(WorkInfo.State.FAILED);
+                return;
+            } else {
+                // TransferIdOrError.Operation.UPLOAD_DOWNLOAD
```
bsiegel, line 312: Unclear what this comment is for

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
bsiegel: Probably just add a clarifying comment. I'm sure this question will come up again in the future otherwise.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdOrError.java
bsiegel: Seeing the changes here later on. Still think we should make this a more generic name, but maybe not an actual `generic` type. Perhaps something like `TransferOperationResult`?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
anuchandy: it's uplifted from WorkManager `cancel` API doc. WorkManager::Worker::cancel() is best effort.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveDataCache.java
anuchandy: I see, yea `while` is more cleaner/readable, will switch to it.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdInfoLiveData.java
anuchandy: I mean, this code path hit for upload or download not for resume. I will make it more clear in the doc instead of just putting enum.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferIdOrError.java
anuchandy: yes, I was thinking the same, generic type is something we can think if we encounter more scenarios where we need to pass value-error through LiveData. `TransferOperationResult` looks good to me.

PR #229, opened 2020-04-27
Enable Check Enforcer

mitchdenny:
This PR sets up the Check Enforcer config on the dev branch.

### DIFF
--- a/eng/CHECKENFORCER
+++ b/eng/CHECKENFORCER
@@ -1,0 +1,2 @@
+format: v0.1-alpha
+minimumCheckRuns: 2


eng/CHECKENFORCER
```diff
@@ -0,0 +1,2 @@
+format: v0.1-alpha
```
weshaggard, line 1: Does checkenforcer work for non-master branches?

mitchdenny: Yes ... it does, but it only reads its configuration from master.

Issue #134, opened 2019-11-21
Implement token credentials policy & interface


bsiegel:
- [ ] Create generic credential types which can be obtained from different providers (bearer token, SAS URL)
- [ ] Create policies to inject the credential into the request

Need to understand more about the auth scenarios we will be supporting on mobile.

PR #238, opened 2020-05-21
Revisiting identity-related classes.

vcolin7:
- Moved AccessToken from sample-app to azure-core. and created its test class.,
- Renamed storage.blob.credentials to storage.blob.credential.
- Made slight corrections to the TokenRequest classes' Javadoc in sample-app.


vcolin7:
@bsiegel @anuchandy PTAL :)

bsiegel:
No problems with the changes so far, but I think ultimately we will want all of the TokenRequest* code to move into the "identity" library too. Basically anything we don't expect developers should have to write themselves, if we can provide it for them, we should. With MSAL that's unfortunately less than we would like, since there is a lot of glue needed to wire up MSAL to the actual application that's using it. But even there we can contribute some of that boilerplate and provide a good sample for the developers to follow in order to implement the remainder of it themselves.

vcolin7:
> 
> 
> No problems with the changes so far, but I think ultimately we will want all of the TokenRequest* code to move into the "identity" library too. Basically anything we don't expect developers should have to write themselves, if we can provide it for them, we should. With MSAL that's unfortunately less than we would like, since there is a lot of glue needed to wire up MSAL to the actual application that's using it. But even there we can contribute some of that boilerplate and provide a good sample for the developers to follow in order to implement the remainder of it themselves.

I agree we will want to move this to an Identity library, do you think that is something we want to do for this preview?

bsiegel:
> I agree we will want to move this to an Identity library, do you think that is something we want to do for this preview?

I think we should just pull it into the identity library now. I was going to suggest that we can just pull it into core if identity hasn't been created yet, but honestly the process of making the identity library is pretty trivial, and there's not much extra code that needs to be pulled into it. Plus I think we would need to do this work for preview 1 anyway.

vcolin7:
@bsiegel @anuchandy I moved our generic credential classes to azure-core as previously discussed and wrote unit tests for all of them. I has special trouble deciding how to approach testing the `okenRequestObservableAuthInterceptor` because `TokenRequestObservable` only exposes LiveData's `observe(LifecycleOwner, TokenRequestObserver)` and we cannot use classes from the Android API like LifecycleOwner in unit tests. I found three different ways to approach this:

- **Use Robolectric** to access Android API classes in unit tests. This is specially slow and would not recommend pulling this dependency unless we had many cases where our unit tests benefited from this. [Commit link](https://github.com/Azure/azure-sdk-for-android/pull/238/commits/0270f5924cc9b9f730321f813eb54397c92f5094).
- **Use Mockito** to mock a `LifecyceOwner` object. This is way faster but still pulls a test dependency on Mockito. [Commit link](https://github.com/Azure/azure-sdk-for-android/pull/238/commits/7a31652ba7a7d1da85f82afdbd126a0a9a87cceb).
- **Add `observeForever(TokenRequestObserver)` and `removeObserver(TokenRequestObserver)` to `TokenRequestObservable`**. This has the benefit of not using any additional test dependencies, however, do we want to expose such functionality to users? If not, I'm not sure it would be a good practice to add package private methods to only be used by unit tests. [Commit link](https://github.com/Azure/azure-sdk-for-android/pull/238/commits/26894c1c4246b40a537c7be7b442bf98b614c528).

bsiegel:
Good questions. I'm curious, since the type `androidx.lifecycle.LifecycleOwner` is in Jetpack rather than in the android OS, is it really true that we can only use this type in a connectedTest?

If that's the case, then arguably the "correct" thing would be to hide the type behind an interface, since the current state means our `TokenRequestObservableAuthInterceptor` is strongly coupled directly to the Android OS. However, I'm not clear how much noise that would end up generating.

For the third option you provided, is it possible to make those added methods package-private? If they are required to be fully public then I would say option three is not viable.

vcolin7:
@bsiegel Thank you for taking the time to read, there are definitely more ways we could go about solving this problem like the ones you suggested.

I spent some time thinking about your question of `LifecycleOwner` being a part of Jetpack and not the Android OS itself and it turns out you are right, I must have been fixed on thinking only UI components such as `Activity` or `Fragment` could make use of `Lifecycle` and `LifecycleOwner`. With that being said, I think the solution of mocking the `LifecycleOwner` ourselves is both elegant and simple, so I will be suing it in this case, thanks a lot! :D

sdk/core/azure-core/src/main/java/com/azure/android/core/credential/TokenRequestHandle.java
```diff
@@ -0,0 +1,136 @@
...
+
+        if (this.error != null) {
+            throw this.error;
+        } else {
+            return this.token;
+        }
+    }
+
+    /**
+     * PACKAGE PRIVATE METHOD.
```
bsiegel, line 80: If these are package-private, why are we just using a comment rather than declaring them with package-private visibility (that is, no visibility modifier)? I know the class itself is package-private, but then why call out certain methods as package-private and others not?

sdk/core/azure-core/src/main/java/com/azure/android/core/credential/TokenRequestObservableAuthInterceptor.java
```diff
@@ -28,30 +28,31 @@
...
     }
 
+    @NonNull
     @Override
-    public Response intercept(Chain chain) throws IOException {
-        if (this.accessToken != null && !this.accessToken.isExpired()) {
-            return setAuthenticationHeader(chain, this.accessToken);
-        } else {
+    public Response intercept(@NonNull Chain chain) throws IOException {
+        if (this.accessToken == null || this.accessToken.isExpired()) {
```
bsiegel, line 50: I know you didn't actually change this code, but in iOS we don't check if the token is expired, we just let the service call fail with the appropriate error response. This goes with the pattern of not implementing logic in the client library, that is also implemented on the service side (since you can never guarantee the same logic will be implemented identically in both places).

Can or should we remove this check? If the answer is yes, it doesn't need to be done in this PR, we should just file an issue.

sdk/core/azure-core/src/main/java/com/azure/android/core/credential/AccessToken.java
```diff
@@ -14,30 +14,30 @@
...
     /**
      * Creates an access token instance.
-     * @param token the token string.
-     * @param expiresAt the expiration time.
+     * @param token The token string.
+     * @param expiresAt The expiration time.
      */
     public AccessToken(String token, OffsetDateTime expiresAt) {
         this.token = token;
         this.expiresAt = expiresAt.minusMinutes(2); // 2 minutes before token expires
```
bsiegel, line 22: I know we talked about using this "T-2 minutes" value as the expiration before, but I couldn't for the life of me figure out where that idea originated. I ended up taking it out of the iOS implementation. If we keep `expiresAt` here (which we may not depending on the decision about checking for expiration), we should probably remove this magic 2 minute adjustment to the expiration time.

vcolin7: I just talked to Jianghao on the Java team about this since they use the 2 minutes subtraction. He explained that the idea is that there's delay between when an access token's expiry time is checked in the `BearerTokenPolicy` and when the request reaches the service, this is more so when the Netty request queue is backed up and a request is sitting there for a long time. However, they are changing this behavior soon so `AccessToken` returns the actual `expiresAt` value and moving it to `SimpleaTokenCache`.

Based on this. I believe we can do as you suggest and remove the subtraction altogether and worry about the delay between a request being made and it being serviced somewhere else.

bsiegel: Interesting, so on iOS we don't look at the expiration at all, we just send the token to the service and let it respond with an error if it's expired. Also we don't currently have any mechanism for caching tokens (which I believe is also true for Android? - correct me if I'm wrong. I will have to look into this caching mechanism, because the logic for caching tokens seemed pretty complex when I started thinking about it.

sdk/core/azure-core/src/main/java/com/azure/android/core/credential/TokenRequestObservableAuthInterceptor.java
vcolin7: Checking for expiration is simple enough with this `AccessToken` class and I believe that we should not make additional calls if we can easily know if the token is still valid, however, I do not know what other checks might be ocurring in the service side, so discussing this a little further sounds like a good idea.

sdk/core/azure-core/src/main/java/com/azure/android/core/credential/TokenRequestHandle.java
vcolin7: Good catch! This comes from the time these classes were originally written and I must have missed it when migrating them to azure-core.

sdk/core/azure-core/src/main/java/com/azure/android/core/credential/AccessToken.java
vcolin7: I agree that checking for expiration in the cache makes way more sense than to do it in the token itself. For now, I think removing this check from the token and let the service tell us if it's expired is the best way to go, like you suggested.

bsiegel:
Does this complete #134?

sdk/core/azure-core/src/test/java/com/azure/android/core/credential/AccessTokenTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.Test;
+import org.threeten.bp.OffsetDateTime;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class AccessTokenTest {
+    private String token = "testToken";
+    private OffsetDateTime time; // Expiration is set to 2 minutes before the given time.
```
bsiegel, line 12: This is no longer true, does this test file need to be updated?

vcolin7: I thought I had pushed the change, let me do that right now.

bsiegel:
:shipit:

PR #239, opened 2020-05-23
Created azure-storage unit tests (excluding TransferManager). Modified public API.

vcolin7:

Partially addresses #189.

- Created unit tests for azure-storage interceptors.
- Created unit tests for StorageBlobClient.
- Polished the StorageBlobClient public API in to be more consistent.


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -921,15 +1015,19 @@ public void onFailure(@NonNull Call<ResponseBody> call, @NonNull Throwable t) {
...
 
                     return new CallAndOptionalResult<>(call, result);
-                } else { // Error response
-                    String strContent = readAsString(response.body());
+                } else {
+                    ResponseBody responseBody = response.body();
+                    String message = responseBody == null ? "no message" : "message: " + readAsString(responseBody);
 
-                    throw new BlobStorageException(strContent, response.raw());
+                    throw new BlobStorageException("Response failed with code: " + response.code() + " and " + message,
```
anuchandy, line 1022: Refer above comment i.e. I would suggest keeping the message as it is. Also the HTTP response.code() is often 200 with the actual error-code exists in the response-header, so this extra text additions may not add much value but make it bit difficult to parse like I mentioned above.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/SasTokenCredentialInterceptor.java
```diff
@@ -28,17 +28,20 @@ public SasTokenCredentialInterceptor(SasTokenCredential credential) {
...
             : sasToken;
 
+        // SAS token is already encoded so its safe to append it to the encoded query from source request.
+        String encodedQuery = requestURL.encodedQuery();
+
+        if (isNullOrEmpty(encodedQuery)) {
+            encodedQuery = sasToken;
+        } else {
+            encodedQuery += "&" + sasToken;
+        }
```
anuchandy, line 43: good catch and fix 👍

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -392,28 +399,28 @@ public ServiceCall downloadWithHeaders(String containerName,
             callback);
     }
 
-    public void stageBlock(String containerName,
+    public BlockBlobsStageBlockResponse stageBlock(String containerName,
```
anuchandy, line 402: The methods in this category are those with "simplest" signature, which should not return any `Response` type but instead returns the actual value in the response, in this case it should be `void`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -392,28 +399,28 @@ public ServiceCall downloadWithHeaders(String containerName,
...
     }
 
     public ServiceCall stageBlock(String containerName,
                                   String blobName,
                                   String base64BlockId,
-                                  byte[] body,
+                                  byte[] blockContent,
                                   byte[] contentMd5,
-                                  Callback<Void> callback) {
+                                  Callback<BlockBlobsStageBlockResponse> callback) {
```
anuchandy, line 419: ref the previous comment, this is the async variant with "simplest" signature, and should be producing value in the Response (i.e. void) not the Response.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
vcolin7: You are right, I must have missed this for `stageBlock`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
vcolin7: So we would make this accept a `Callback<Void>` instead, right?

anuchandy: right, I see you already made this change, sorry for delay.

sdk/core/azure-core/src/main/java/com/azure/android/core/internal/util/test/EnqueueMockResponse.java
```diff
@@ -1,4 +1,4 @@
-package com.azure.android.core.http.interceptor;
+package com.azure.android.core.internal.util.test;
```
anuchandy, line 1: I guess there are two things in the `test` project:

1. Unit tests for types in the `core` project.
2. Test types that we want to share with other "dependent" projects (e.g., storage).

For the second category, does it make sense to keeping these shared types rooted at `com.azure.android.core.test,` hence consistent with the established namespace in java [sdk](https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/core/azure-core-test/src/main/java/com/azure/core/test). Its a shippable package in java SDK, we haven't there yet for android but agreeing on the namespace will be a good start. Thoughts?

anuchandy: Between why did we move "shared test" types from "src/test" to "src/main"?

anuchandy: If discoverability of types are the reason for "src/test" -> "src/main" then worth checking this: https://stackoverflow.com/questions/5644011/multi-project-test-dependencies-with-gradle/60138176#60138176

anuchandy:
@vcolin7 should we merge this?

vcolin7:
@anuchandy Let's merge this for now and we will revisit the strategy to include core test classes at a later point.

PR #241, opened 2020-06-09
Transfer apis in storage client

anuchandy:
This PR updates the `StorageBlobClient` to expose transfer APIs (upload, download, pause, resume, cancel).

The `StorageClient.Builder` now allows setting the transfer constraints. Below shows how to create `StorageBlobClient`.

```java
StorageBlobClient client1 = new StorageBlobClient
    .Builder("com.azure.android.storage.sample")
        .setBlobServiceUrl(storageConfiguration.getBlobServiceUrl())
        .setRequiredNetworkType(NetworkType.CONNECTED)
        .build();
```
Note that the builder constructor now takes the "storage client id" that uniquely identifies the `StorageBlobClient` this builder creates. As part of executing `build()` it adds the created `StorageBlobClient` to the global map.

We continue to support creating a new `StorageBlobClient` from an existing client:

```java
StorageBlobClient client2 = client1
    .newBuilder("com.azure.android.storage.sample.upload")
        .setCredentialInterceptor(authInterceptor)
        .setRequiresBatteryNotLow(true)
        .build();
```

`TransferClient` has been updated to have a singleton instance. It is retrieve as: 

```java
TransferClient transferClient = TransferClient.getInstance(context);    
```

This is similar to the way we get WorkManager:

```java
WorkManager workManager = WorkManager.getInstance(context);
```

A `TransferClient` instance will not hold a reference to the context. Some read on the subject tells its a bad idea to keep context, so, like `WorkManager`, `TransferClient` uses the context for initialization and release it.

Some notes on the PR:

1. We want to hide `TransferClient` for preview1, which means `TransferClient` and `StorageBlobClient` may have to exist in the same package. This PR won't move files around for this; just want this PR to highlight the concept, logic, API changes for easy review.

2. There are two new types `UploadRequest`|`DownloadRequest`; it is easy to merge these two but didn't do it - just following the existing design of separate internal types for upload|download. Also, these types are anyway going to be internal once `#1` is addressed.

PR #242, opened 2020-06-09
Javadoc updates after local review of generated docs.

anuchandy:
Copying over the JavaDoc for the stage, commit APIs from Java SDK (swagger) to StorageBlobClient

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RequestIdInterceptor.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RequestIdInterceptor.java
@@ -20,9 +20,9 @@ import okhttp3.Response;
  */
 public class RequestIdInterceptor implements Interceptor {
     /**
-     * Intercept the current request in the pipeline and apply x-ms-client-request-id header.
-     *
-     * @param chain provide access to the request to apply the x-ms-client-request-id header.
+     * Intercept the current request in the pipeline and apply the "x-ms-client-request-id" header.
+
+     * @param chain Provide access to the request to apply the "x-ms-client-request-id" header.
      *
      * @return Response from the next interceptor in the pipeline.
      * @throws IOException If an IO error occurs while processing the request and response.
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RequestIdInterceptor.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RequestIdInterceptor.java
@@ -19,6 +19,14 @@ import okhttp3.Response;
  * the request.
  */
 public class RequestIdInterceptor implements Interceptor {
+    /**
+     * Intercept the current request in the pipeline and apply x-ms-client-request-id header.
+     *
+     * @param chain provide access to the request to apply the x-ms-client-request-id header.
+     *
+     * @return Response from the next interceptor in the pipeline.
+     * @throws IOException If an IO error occurs while processing the request and response.
+     */
     @NonNull
     @Override
     public Response intercept(@NonNull Chain chain) throws IOException {
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
@@ -57,14 +57,19 @@ public interface ClientLogger {
      */
     int LOG_LEVEL_ERROR = 4;
 
+
     /**
-     * Returns the logger's log level.
+     * Get the current log level.
+     *
+     * @return the current log level
      */
     @LogLevel
     int getLogLevel();
 
     /**
      * Sets the log level for the logger.
+     *
+     * @param logLevel the log level
      */
     void setLogLevel(@LogLevel int logLevel);
 
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
@@ -59,9 +59,9 @@ public interface ClientLogger {
 
 
     /**
-     * Get the current log level.
+     * Gets the current log level.
      *
-     * @return the current log level
+     * @return The current log level.
      */
     @LogLevel
     int getLogLevel();
@@ -69,7 +69,7 @@ public interface ClientLogger {
     /**
      * Sets the log level for the logger.
      *
-     * @param logLevel the log level
+     * @param logLevel The log level.
      */
     void setLogLevel(@LogLevel int logLevel);
 
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
@@ -3,13 +3,15 @@
 
 package com.azure.android.storage.blob;
 
+import android.net.Uri;
+
 import com.azure.android.core.http.Callback;
 import com.azure.android.core.http.ServiceCall;
 import com.azure.android.core.http.ServiceClient;
 import com.azure.android.core.http.interceptor.AddDateInterceptor;
 import com.azure.android.core.internal.util.serializer.SerializerFormat;
 import com.azure.android.storage.blob.models.AccessTier;
-import com.azure.android.storage.blob.models.BlobDownloadAsyncResponse;
+import com.azure.android.storage.blob.models.BlobDownloadResponse;
 import com.azure.android.storage.blob.models.BlobGetPropertiesHeaders;
 import com.azure.android.storage.blob.models.BlobHttpHeaders;
 import com.azure.android.storage.blob.models.BlobItem;
@@ -24,6 +26,7 @@ import com.azure.android.storage.blob.models.CpkInfo;
 import com.azure.android.storage.blob.models.ListBlobsIncludeItem;
 import com.azure.android.storage.blob.models.ListBlobsOptions;
 
+import java.io.File;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -33,6 +36,9 @@ import okhttp3.ResponseBody;
 
 /**
  * Client for Storage Blob service.
+ *
+ * <p>
+ * This client is instantiated through {@link StorageBlobClient.Builder}.
  */
 public class StorageBlobClient {
     private final ServiceClient serviceClient;
@@ -160,10 +166,11 @@ public class StorageBlobClient {
     }
 
     /**
-     * Reads the blob's metadata & properties.
+     * Reads the blob's metadata and properties.
      *
      * @param containerName The container name.
      * @param blobName      The blob name.
+     * @return the blob's metadata and properties
      */
     public BlobGetPropertiesHeaders getBlobProperties(String containerName,
                                                       String blobName) {
@@ -171,7 +178,7 @@ public class StorageBlobClient {
     }
 
     /**
-     * Reads the blob's metadata & properties.
+     * Reads the blob's metadata and properties.
      *
      * @param containerName The container name.
      * @param blobName      The blob name.
@@ -187,7 +194,7 @@ public class StorageBlobClient {
     }
 
     /**
-     * Reads a blob's metadata & properties.
+     * Reads a blob's metadata and properties.
      *
      * @param containerName         The container name.
      * @param blobName              The blob name.
@@ -226,7 +233,7 @@ public class StorageBlobClient {
     }
 
     /**
-     * Reads a blob's metadata & properties.
+     * Reads a blob's metadata and properties.
      *
      * @param containerName         The container name.
      * @param blobName              The blob name.
@@ -242,6 +249,7 @@ public class StorageBlobClient {
      * @param requestId             Provides a client-generated, opaque value with a 1 KB character limit that is
      *                              recorded in the analytics logs when storage analytics logging is enabled.
      * @param cpkInfo               Additional parameters for the operation.
+     *  @param callback             Callback that receives the response.
      * @return The service call object, representing the request scheduled for execution.
      */
     public ServiceCall getBlobPropertiesWithRestResponse(String containerName,
@@ -269,11 +277,19 @@ public class StorageBlobClient {
     /**
      * Reads the entire blob.
      *
+     * <p>
+     * This method will execute a raw HTTP GET in order to download a single blob to the destination.
+     * It is **STRONGLY** recommended that you use the {@link com.azure.android.storage.blob.transfer.TransferClient#download(String, String, String, File)}
+     * or {@link com.azure.android.storage.blob.transfer.TransferClient#download(String, String, String, Uri)} method instead - that method will
+     * manage the transfer in the face of changing network conditions, and is able to transfer multiple
+     * blocks in parallel.
+     *`
      * @param containerName The container name.
      * @param blobName      The blob name.
+     * @return the response containing blob bytes
      */
-    public ResponseBody download(String containerName,
-                                 String blobName) {
+    public ResponseBody rawDownload(String containerName,
+                                    String blobName) {
         return storageBlobServiceClient.download(containerName,
             blobName);
     }
@@ -281,14 +297,21 @@ public class StorageBlobClient {
     /**
      * Reads the entire blob.
      *
+     * <p>
+     * This method will execute a raw HTTP GET in order to download a single blob to the destination.
+     * It is **STRONGLY** recommended that you use the {@link com.azure.android.storage.blob.transfer.TransferClient#download(String, String, String, File)}
+     * or {@link com.azure.android.storage.blob.transfer.TransferClient#download(String, String, String, Uri)} method instead - that method will
+     * manage the transfer in the face of changing network conditions, and is able to transfer multiple
+     * blocks in parallel.
+     *
      * @param containerName The container name.
      * @param blobName      The blob name.
      * @param callback      Callback that receives the response.
      * @return The service call object, representing the request scheduled for execution.
      */
-    public ServiceCall download(String containerName,
-                                String blobName,
-                                Callback<ResponseBody> callback) {
+    public ServiceCall rawDownload(String containerName,
+                                   String blobName,
+                                   Callback<ResponseBody> callback) {
         return storageBlobServiceClient.download(containerName,
             blobName,
             callback);
@@ -296,6 +319,13 @@ public class StorageBlobClient {
 
     /**
      * Reads a range of bytes from a blob.
+     *
+     * <p>
+     * This method will execute a raw HTTP GET in order to download a single blob to the destination.
+     * It is **STRONGLY** recommended that you use the {@link com.azure.android.storage.blob.transfer.TransferClient#download(String, String, String, File)}
+     * or {@link com.azure.android.storage.blob.transfer.TransferClient#download(String, String, String, Uri)} method instead - that method will
+     * manage the transfer in the face of changing network conditions, and is able to transfer multiple
+     * blocks in parallel.
      *
      * @param containerName         The container name.
      * @param blobName              The blob name.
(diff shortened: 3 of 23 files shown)


anuchandy:
@vcolin7 thanks for the suggestions, could you do another round of review? I did update couple of more files.

anuchandy:
JavaDoc based on this PR: https://anuchandy.github.io/android-sdk-doc-review/index.html

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RequestIdInterceptor.java
```diff
@@ -19,6 +19,14 @@
  * the request.
  */
 public class RequestIdInterceptor implements Interceptor {
+    /**
+     * Intercept the current request in the pipeline and apply x-ms-client-request-id header.
+     *
+     * @param chain provide access to the request to apply the x-ms-client-request-id header.
```
vcolin7, line 25:
```suggestion
     * Intercept the current request in the pipeline and apply the "x-ms-client-request-id" header.

     * @param chain Provide access to the request to apply the "x-ms-client-request-id" header.
```

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -57,14 +57,19 @@ static ClientLogger getDefault(String tag) {
      */
     int LOG_LEVEL_ERROR = 4;
 
+
     /**
-     * Returns the logger's log level.
+     * Get the current log level.
+     *
+     * @return the current log level
```
vcolin7, line 64:
```suggestion
     * Gets the current log level.
     *
     * @return The current log level.
```

sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java
```diff
@@ -57,14 +57,19 @@ static ClientLogger getDefault(String tag) {
...
      */
     @LogLevel
     int getLogLevel();
 
     /**
      * Sets the log level for the logger.
+     *
+     * @param logLevel the log level
```
vcolin7, line 72:
```suggestion
     * @param logLevel The log level.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/models/BlobStorageException.java
```diff
@@ -41,21 +41,27 @@ public BlobStorageException(String message, okhttp3.Response response) {
...
+     *
+     * @return The message.
      */
     public String getServiceMessage() {
         return super.getMessage();
     }
 
     /**
-     * @return The status code on the response.
+     * Gets the status code of the response error.
```
vcolin7, line 62:
```suggestion
     * Gets the status code of the error response.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/SasTokenCredentialInterceptor.java
```diff
@@ -15,7 +15,7 @@
 import static com.azure.android.core.util.CoreUtil.isNullOrEmpty;
 
 /**
- *  Interceptor that append SAS token to the request Uri.
+ *  Pipeline interceptor that append SAS token to the request Uri.
```
vcolin7, line 18:
```suggestion
 *  Pipeline interceptor that append SAS token to the request URI.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/SasTokenCredentialInterceptor.java
```diff
@@ -24,6 +24,14 @@ public SasTokenCredentialInterceptor(SasTokenCredential credential) {
         this.credential = credential;
     }
 
+    /**
+     * Intercept the current request in the pipeline and apply the SAS token.
+     *
+     * @param chain Provide access to the request to apply the SAS token.
+     *
+     * @return Response from the next interceptor in the pipeline.
```
vcolin7, line 32:
```suggestion
     * @return Response From the next interceptor in the pipeline.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/ResponseHeadersValidationInterceptor.java
```diff
@@ -60,6 +67,14 @@ public ResponseHeadersValidationInterceptor(Collection<String> headerNames, Clie
         logger = clientLogger;
     }
 
+    /**
+     * Intercept and validate that a collection of headers have consistent values between a request and a response.
+     *
+     * @param chain Provide access to the response to be validated.
+     *
+     * @return Response from the next interceptor in the pipeline.
```
vcolin7, line 75:
```suggestion
     * @return Response From the next interceptor in the pipeline.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/ResponseHeadersValidationInterceptor.java
```diff
@@ -52,6 +56,9 @@ public ResponseHeadersValidationInterceptor(Collection<String> headerNames) {
     /**
      * Constructor that accepts a list of header names to validate. Adds two mandatory Storage header names as well
      * and uses a provided {@link ClientLogger}.
+     *
+     * @param headerNames  the header names
+     * @param clientLogger the logger
```
vcolin7, line 61:
```suggestion
     * @param headerNames The header names.
     * @param clientLogger The logger.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/ResponseHeadersValidationInterceptor.java
```diff
@@ -44,6 +46,8 @@ public ResponseHeadersValidationInterceptor(ClientLogger clientLogger) {
     /**
      * Constructor that accepts a list of header names to validate. Adds two mandatory Storage header names as well.
      * and uses a default {@link ClientLogger}.
+     *
+     * @param headerNames the header names
```
vcolin7, line 50:
```suggestion
     * @param headerNames The header names.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/ResponseHeadersValidationInterceptor.java
```diff
@@ -34,6 +34,8 @@ public ResponseHeadersValidationInterceptor() {
 
     /**
      * Constructor that adds two mandatory headers used by Storage and uses a provided {@link ClientLogger}.
+     *
+     * @param clientLogger the logger
```
vcolin7, line 38:
```suggestion
     * @param clientLogger The logger.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/NormalizeEtagInterceptor.java
```diff
@@ -10,16 +10,22 @@
...
+     * <p>
      * The service is inconsistent in whether or not the ETag header value has quotes. This method will check if the
      * response returns an ETag value, and if it does, remove any quotes that may be present to give the user a more
      * predictable format to work with.
      *
-     * @return An updated response with post processing steps applied.
+     * @param chain Provide access to the response containing ETag header to normalize.
+     *
+     * @return Response from the next interceptor in the pipeline.
+     * @throws IOException If an IO error occurs while processing the request and response
```
vcolin7, line 28:
```suggestion
     * @return Response From the next interceptor in the pipeline.
     * @throws IOException If an IO error occurs while processing the request and response.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -265,25 +277,41 @@ public ServiceCall getBlobPropertiesWithRestResponse(String containerName,
...
+     * manage the transfer in the face of changing network conditions, and is able to transfer multiple
+     * blocks in parallel.
+     *`
      * @param containerName The container name.
      * @param blobName      The blob name.
+     * @return the response containing blob bytes
      */
-    public ResponseBody download(String containerName,
-                                 String blobName) {
+    public ResponseBody rawDownload(String containerName,
```
vcolin7, line 291: What is the reason behind renaming this to `rawDownload`? Would `download` be reserved for the `TransferClient`? Is the word `raw` something commonly used in the Java world for an operation that directly calls an endpoint or something like that?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -265,25 +277,41 @@ public ServiceCall getBlobPropertiesWithRestResponse(String containerName,
...
+     *
      * @param containerName The container name.
      * @param blobName      The blob name.
      * @param callback      Callback that receives the response.
+     * @return The service call object, representing the request scheduled for execution.
      */
-    public ServiceCall download(String containerName,
-                                String blobName,
-                                Callback<ResponseBody> callback) {
+    public ServiceCall rawDownload(String containerName,
```
vcolin7, line 312: Same comment about renaming this method.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -239,6 +249,8 @@ public BlobGetPropertiesResponse getBlobPropertiesWithRestResponse(String contai
      * @param requestId             Provides a client-generated, opaque value with a 1 KB character limit that is
      *                              recorded in the analytics logs when storage analytics logging is enabled.
      * @param cpkInfo               Additional parameters for the operation.
+     *  @param callback             Callback that receives the response.
```
vcolin7, line 252:
```suggestion
     * @param callback             Callback that receives the response.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -265,25 +277,41 @@ public ServiceCall getBlobPropertiesWithRestResponse(String containerName,
...
+     * <p>
+     * This method will execute a raw HTTP GET in order to download a single blob to the destination.
+     * It is **STRONGLY** recommended that you use the {@link com.azure.android.storage.blob.transfer.TransferClient#download(String, String, String, File)}
+     * or {@link com.azure.android.storage.blob.transfer.TransferClient#download(String, String, String, Uri)} method instead - that method will
+     * manage the transfer in the face of changing network conditions, and is able to transfer multiple
+     * blocks in parallel.
+     *`
      * @param containerName The container name.
      * @param blobName      The blob name.
+     * @return the response containing blob bytes
```
vcolin7, line 289:
```suggestion
     * @return The response containing the blob's bytes.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -313,17 +348,17 @@ public ServiceCall download(String containerName,
...
-                                                              String snapshot,
-                                                              Integer timeout,
-                                                              BlobRange range,
-                                                              BlobRequestConditions blobRequestConditions,
-                                                              Boolean getRangeContentMd5,
-                                                              Boolean getRangeContentCrc64,
-                                                              String version,
-                                                              String requestId,
-                                                              CpkInfo cpkInfo) {
+    public BlobDownloadResponse rawDownloadWithRestResponse(String containerName,
```
vcolin7, line 351: Same comment about renaming `download`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -365,19 +407,21 @@ public BlobDownloadAsyncResponse downloadWithRestResponse(String containerName,
...
-                                                Integer timeout,
-                                                BlobRange range,
-                                                BlobRequestConditions blobRequestConditions,
-                                                Boolean getRangeContentMd5,
-                                                Boolean getRangeContentCrc64,
-                                                String version,
-                                                String requestId,
-                                                CpkInfo cpkInfo,
-                                                Callback<BlobDownloadAsyncResponse> callback) {
+    public ServiceCall rawDownloadWithRestResponse(String containerName,
```
vcolin7, line 413: Same comment about renaming `download`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -461,16 +568,30 @@ public ServiceCall stageBlockWithRestResponse(String containerName,
...
+     * {@link StorageBlobClient#stageBlock(String, String, String, byte[], byte[])} operation. You can call commit Block List
+     * to update a blob by uploading only those blocks that have changed, then committing the new and existing blocks together.
+     * You can do this by specifying whether to commit a block from the committed block list or from the uncommitted block list,
+     * or to commit the most recently uploaded version of the block, whichever list it may belong to.
+     *
+     * @param containerName  The container name.
+     * @param blobName       The blob name.
+     * @param base64BlockIds The block IDs.
+     * @param overwrite      Indicate whether to overwrite the block list if already exists.
+     * @return the properties of the block blob
```
vcolin7, line 593:
```suggestion
     * @return The properties of the block blob
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -210,10 +210,10 @@ private TransferClient(Context context,
      * @param storageBlobClientId the identifier of the blob storage client to use for the download
```
vcolin7, line 210:
```suggestion
     * @param storageBlobClientId The identifier of the blob storage client to use for the download
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -607,6 +607,8 @@ private StopCheck(boolean canStop, boolean isUpload) {
 
         /**
          * Create a new {@link TransferClient} builder.
+         *
+         * @param context the application context
```
vcolin7, line 611:
```suggestion
         * @param context The application context.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/TransferClient.java
```diff
@@ -623,8 +625,8 @@ public Builder(@NonNull Context context) {
          *   it is recommended to use a value unique to your application (e.g. "MyApplication").
          *   If your application uses multiple clients with different configurations, use a value unique
          *   to both your application and the configuration (e.g. "MyApplication.userClient").
-         * @param storageBlobClient the blob storage client
-         * @return Builder with the provided blob storage client set
+         * @param storageBlobClient the blob storage client.
```
vcolin7, line 628:
```suggestion
         * @param storageBlobClient The blob storage client.
```

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/models/BlobDownloadResponse.java
```diff
@@ -12,21 +12,21 @@
 /**
  * This class contains the response information returned from the server when downloading a blob.
  */
-public final class BlobDownloadAsyncResponse extends ResponseBase<BlobDownloadHeaders, ResponseBody> {
+public final class BlobDownloadResponse extends ResponseBase<BlobDownloadHeaders, ResponseBody> {
```
vcolin7, line 15: Java has two distinct types: [`BlobDownloadResponse`](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobDownloadResponse.java) and [`BlobDownloadAsyncResponse`](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/models/BlobDownloadAsyncResponse.java). I originally decided to go for the latter because our use case lined up with what Java has.

Do you think we will add something like Java's `BlobDownloadResponse` where no ResponseBody is returned but rather written to something like an `OutputStream`? I tend to think so.

anuchandy: Right, Java has `BlobDownloadAsyncResponse` to support `Flux` that asynchronously streams the bytes from the connection, so the text "async" in the model indicates that it composes "async" Flux.

In the Android case, `BlobDownloadAsyncResponse`  is composing `ResponseBody` that exposes `InputStream` from which we can "synch"-ronously read the bytes from the connection. This is why I removed the text, in addition to that, it looks also weird when that type listed in the JavaDoc with our other Response types.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
anuchandy: It was a preparation for the work of merging `TransferClient` APIs to `StorageClient`, similar to what iOS did, i.e. we want to discourage users from using raw GET APIs and encourage to use "managed" download. I'm not completely happy with this name though.

vcolin7: Would it be a bad idea to keep `download` for both clients but to encourage people to use the managed operation both in the documentation and code samples? Any thoughts on this @bsiegel?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/models/BlobDownloadResponse.java
vcolin7: Since the decision to read the bytes synchronously would come down to us, do you think there could be a point where we would want to go the fully async route with `ResponseBody`? (i.e. using [`byteStream`](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response-body/byte-stream/)).

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
vcolin7: Oh nevermind, looking at your other PR (#241) I piucked up on what you mean by merging both clients. I think having two different names is good, while I also feel there might be something better than the `raw` prefix.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/models/BlobDownloadResponse.java
anuchandy: `byteStream()` API returns "InputStream", in my previous comment when I said "sync" read I mean that the InputStream::read API itself is blocking. i.e. can't do something like below with IS:

```java
InputStream.readAsync(bytes_count, new CallBack(byte[] bytes) {
});
```

So even though we can asynchronously return `BlobDownloadResponse`, we can't asynchronously read the content:

```java
downloadAsync(params, new CallBack<BlobDownloadResponse>() {
    // This is async
    public void onResponse(BlobDownloadResponse res) {
         // below read is not async but sync.
         res.resposeBody.byteStream().read();
    }
}
```

vcolin7: Does this mean we won't be able to asynchronously read content if we are using OkHttp?

anuchandy: yes, OkHttp is natively built on blocking socket not based on selectable channel(e.g. netty). But it should be fine in the mobile space, even if we use native UrlConnection (like android SDK from other cloud providers), it still sync. I think we don't have to worry about it at least now.

anuchandy:
@vcolin7 if you're happy with the changes, can I take this in? I guess the text 'raw' in the download method name is the pending one, we can discuss that as part of the internal doc review and update it later if needed.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/models/BlobDownloadResponse.java
vcolin7: Thanks for the thorough explanation! :)

PR #244, opened 2020-06-13
Added public APIs for deleting blobs.

vcolin7:
Addresses #195.


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -1335,6 +1446,96 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
...
+                    callback.onFailure(t);
+                }
+            });
+
+            return new CallAndOptionalResult<>(call, null);
+        } else {
+            Response<Void> response = executeCall(call);
+
+            if (response.isSuccessful()) {
+                if (response.code() == 200) {
```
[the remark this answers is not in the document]
anuchandy: The java SDK API is [here](https://github.com/Azure/azure-sdk-for-java/blob/4871ff56eec2e0615973a6a32fe09ee2961771f0/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/BlobsImpl.java#L107) for reference.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -1335,6 +1446,96 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
...
+                                                                                   String ifMatch,
+                                                                                   String ifNoneMatch,
+                                                                                   String requestId,
+                                                                                   Callback<BlobDeleteResponse> callback) {
+        DateTimeRfc1123 ifModifiedSinceConverted = ifModifiedSince == null ? null :
+            new DateTimeRfc1123(ifModifiedSince);
+        DateTimeRfc1123 ifUnmodifiedSinceConverted = ifUnmodifiedSince == null ? null :
+            new DateTimeRfc1123(ifUnmodifiedSince);
+
+        Call<Void> call = service.delete(containerName,
```
anuchandy, line 1467: This should be `Call<ResponseBody>` instead of  `Call<Void>`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -1485,5 +1686,19 @@ T getResult() {
                                            @Header("x-ms-encryption-key") String encryptionKey,
                                            @Header("x-ms-encryption-key-sha256") String encryptionKeySha256,
                                            @Header("x-ms-encryption-algorithm") EncryptionAlgorithmType encryptionAlgorithm);
+
+        @DELETE("{containerName}/{blob}")
+        Call<Void> delete(@Path("containerName") String containerName,
```
anuchandy, line 1691: return type should be `Call <ResponseBody>`

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
anuchandy: The reason for "ResponseBody" is, we want later read the body if the delete fails i.e. if service returns non-202.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -1335,6 +1446,96 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
...
+            deleteSnapshots,
+            ifModifiedSinceConverted,
+            ifUnmodifiedSinceConverted,
+            ifMatch,
+            ifNoneMatch,
+            XMS_VERSION, // TODO: Replace with 'version'.
+            requestId);
+
+        if (callback != null) {
+            executeCall(call, new retrofit2.Callback<Void>() {
```
anuchandy, line 1481: `new retrofit2.Callback<ResponseBody>()`

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -1335,6 +1446,96 @@ public void onFailure(Call<ResponseBody> call, Throwable t) {
...
+                public void onResponse(@NonNull Call<Void> call, @NonNull Response<Void> response) {
+                    if (response.isSuccessful()) {
+                        if (response.code() >= 200 && response.code() < 300) {
+                            BlobDeleteHeaders typedHeaders = deserializeHeaders(response.headers(),
+                                BlobDeleteHeaders.class);
+
+                            callback.onResponse(new BlobDeleteResponse(response.raw().request(),
+                                response.code(),
+                                response.headers(),
+                                response.body(),
```
anuchandy, line 1492: since we need to change from `Call<Void>` to `Call<ResponseBody>` here instead of `response.body()` it should be `null`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -547,6 +551,161 @@ public ServiceCall commitBlockListWithRestResponse(String containerName,
...
     }
 
+    /**
+     * Reads the blob's metadata & properties.
+     *
+     * @param containerName The container name.
+     * @param blobName      The blob name.
+     * @return The blob's metadata.
+     */
+    Void delete(String containerName,
```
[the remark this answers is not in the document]
vcolin7: Thank you, I took the template from `getBlobProperties` and forgot to modify it.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -547,6 +551,161 @@ public ServiceCall commitBlockListWithRestResponse(String containerName,
...
+    ServiceCall delete(String containerName,
+                       String blobName,
+                       Callback<Void> callback) {
+        return storageBlobServiceClient.delete(containerName,
+            blobName,
+            callback);
+    }
+
+    /**
+     * If the storage account's soft delete feature is disabled then, when a blob is deleted, it is permanently
```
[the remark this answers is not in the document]
vcolin7: I think you are right. I'll add a more brief explanation.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
vcolin7: I assume we should also do this for APIs like `getBlobProperties` where we don't expect a body in a successful call, but there could be one in a failed call, right?

vcolin7: Also, the reason I went with `Void` is because the Java API returns a `Mono<Void>` for the [delete operation](https://github.com/Azure/azure-sdk-for-java/blob/712e35f4ad597d43d34c678cc263dbc9095a3ed6/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/specialized/BlobAsyncClientBase.java#L1071) and the [`BlobDeleteResponse`](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/models/BlobsDeleteResponse.java) object.

vcolin7: I guess this means we want to return a `Void` at the public API level but a `ResponseBody` at the service interface level. Is this correct?

vcolin7:
Thanks for the comments @anuchandy. I have addressed them, made a couple corrections for how other we generate error messages for other operations and added an `undelete` API.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
anuchandy: _I assume we should also do this for APIs like getBlobProperties where we don't expect a body in a successful call, but there could be one in a failed call, right?_

Yes, it seems there is a similar bug in `getBlobProperties` as well, that needs to be fixed.

_I guess this means we want to return a Void at the public API level but a ResponseBody at the service interface level. Is this correct?_

right, like java SDK, from public API we return `Void` but in internal service interface we need `ResponseBody` so that we can read error body from it and throw.

anuchandy:
@vcolin7 thank you!. Btw is `undelete` included in the set of preview API we're going to support?

vcolin7:
@anuchandy `undelete` is not in the set of APIs we planned for preview 1, I saw it mentioned in the documentation for the `delete` APi in Java and thought adding it wouldn't hurt, since it's people might want to have the ability to undelete a blob if they can delete one. Any thoughts on this @bsiegel?

anuchandy:
@vcolin7 as discussed in sync-up let's remove `undelete` from this PR so that we can get this PR in. 

The `undelete` can be a new PR and exposing it in preview1 can be discussed with Adrian once he is back.

PR #245, opened 2020-06-16
Storage sample application in Kotlin.

anuchandy:


anuchandy:
Closing this; the sample needs to be rewrked once the SDK is on the top of the new module system.

PR #246, opened 2020-06-23
Added public APIs for undeleting blobs.

vcolin7:


bsiegel:
Overall this looks good - two things:

1.) Do we definitely want this functionality for preview 1? If so we need to also add to the iOS SDK. I'm leaning towards just shelving this until later preview(s).

2.) If we do want to ship for preview 1, I think we had some discussions about what public methods should be offered. IIRC we landed on having two versions of each: the first which takes all required parameters only, and the second which takes all required parameters + an Options object. Regardless of which method is called, the callback should always be required, and the callback will always receive both the result object as well as the full response object (which the customer can always ignore).

vcolin7:
Hey @bsiegel, about points no. 1: I don't think this API is necessary for preview 1, I do however think it is nice to have `undelete` if we have a `delete` API, just in case people delete stuff and want to undo it at a later point in time.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -895,6 +896,104 @@ public ServiceCall commitBlockListWithRestResponse(String containerName,
...
+     * Undelete a blob that was previously soft-deleted.
+     * <p>
+     * Undelete restores the content and metadata of a soft-deleted blob and/or any associated soft-deleted snapshots.
+     *
+     * @param containerName The container name.
+     * @param blobName      The blob name.
+     * @param callback      Callback that receives the response.
+     * @return A handle to the service call.
+     */
+    ServiceCall undelete(String containerName,
```
bsiegel, line 922: This should be in the AsyncClient

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -895,6 +896,104 @@ public ServiceCall commitBlockListWithRestResponse(String containerName,
...
+     * @param url           The URL to the deleted blob.
+     * @param timeout       The timeout parameter is expressed in seconds. For more information, see
+     *                      &lt;a href="https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/setting-timeouts-for-blob-service-operations"&gt;Setting Timeouts for Blob Service Operations.&lt;/a&gt;.
+     * @param requestId     Provides a client-generated, opaque value with a 1 KB character limit that is
+     *                      recorded in the analytics logs when storage analytics logging is enabled.
+     * @param comp          The name of the desired operation to perform. In this case, it should be 'undelete'.
+     * @param callback      Callback that receives the response.
+     * @return A handle to the service call.
+     */
+    ServiceCall undeleteWithResponse(String containerName,
```
bsiegel, line 977: This should be in the AsyncClient

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -562,6 +564,127 @@ ServiceCall stageBlockWithRestResponse(String containerName,
...
+     * @param blobName      The blob name.
+     * @param url           The URL to the deleted blob.
+     * @param timeout       The timeout parameter is expressed in seconds. For more information, see
+     *                      &lt;a href="https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/setting-timeouts-for-blob-service-operations"&gt;Setting Timeouts for Blob Service Operations.&lt;/a&gt;.
+     * @param requestId     Provides a client-generated, opaque value with a 1 KB character limit that is
+     *                      recorded in the analytics logs when storage analytics logging is enabled.
+     * @param comp          The name of the desired operation to perform. In this case, it should be 'undelete'.
+     * @return A response object containing the details of the delete operation.
+     */
+    BlobUndeleteResponse undeleteWithResponse(String containerName,
```
bsiegel, line 634: This should not take "comp" as a parameter. For undelete it must always be the string "undelete".

bsiegel: I suspect the right place to hard-code the value for "comp" is actually in `undeleteWithRestResponseIntern`

vcolin7:
@bsiegel I applied the requested changes and also made sure to adapt the undelete methods to include the use of `CancellationToken`. I will also be closing this PR for now but will push this branch to our Android SDK repo, since right now it only resides in my personal fork.

PR #253, opened 2020-07-08
Adding core RetryInterceptor that honors CancellationToken.cancel(), enables Fixed and Expotential stratgies

anuchandy:


sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RetryInterceptor.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * Intercept any exception in the pipeline or the HTTP response error and retry if recoverable.
+
+     * @param chain Provide access to the response.
+     *
+     * @return Response from the next interceptor in the pipeline.
+     * @throws IOException If the pipeline gets canceled or an IO error that cannot be
+     * retried any more, for example, max retry limit reached.
```
vcolin7, line 80:
```suggestion
     * @throws IOException If the pipeline gets canceled or an there is an IO error that
     * indicates the request cannot be retried any more, for example, max retry limit reached.
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RetryInterceptor.java
```diff
@@ -0,0 +1,214 @@
...
+                response = chain.proceed(request);
+            } catch (Exception e) {
+                exception = e;
+            }
+
+            // Check for cancellation after Proceed.
+            if (chain.call().isCanceled()) {
+                try {
+                    if (exception != null) {
+                        // The later interceptors those executes as a result of above 'chain.proceed' may throw
```
vcolin7, line 109:
```suggestion
                        // The later interceptors that execute as a result of the above 'chain.proceed' may throw
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RetryInterceptor.java
```diff
@@ -0,0 +1,214 @@
...
+                        // IOException("Cancelled") [e.g. okhttp3.internal.http.RetryAndFollowUpInterceptor]
+                        // if it identified that call is cancelled, we don't want to retry on such cases.
```
vcolin7, line 111:
```suggestion
                        // If it is identified that the call is cancelled, we don't want to retry on such cases.
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RetryInterceptor.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * Get an instance of {@link RetryInterceptor} that uses fixed backoff delay retry strategy.
+     *
+     * @param maxRetries The maximum number of times to retry.
+     * @param delay The fixed backoff delay applied before every retry.
+     * @return The retry interceptor.
+     */
+    public static RetryInterceptor fixedDelay(int maxRetries, Duration delay) {
```
vcolin7, line 43: Maybe it is obvious to people that you will get an interceptor with a specific type of delay with these static methods, however, I would change the name just to be a bit more explicit about it. Right now someone that only looks at the method name could think you are getting some sort of delay from this, instead of the interceptor itself.

```suggestion
    public static RetryInterceptor withFixedDelay(int maxRetries, Duration delay) {
```

Not sure this is the best name for it, but I hope you get the idea :)

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RetryInterceptor.java
```diff
@@ -0,0 +1,214 @@
...
+     * Get an instance of {@link RetryInterceptor} that uses a default full jitter backoff
+     * retry strategy.
+     *
+     * <p>
+     * The retry strategy by default retries maximum 3 times, uses 800 milliseconds as
+     * the default base delay and uses 8 seconds as default maximum backoff delay before a retry.
+     *
+     * @return The retry interceptor.
+     */
+    public static RetryInterceptor exponentialBackoff() {
```
vcolin7, line 57:
```suggestion
    public static RetryInterceptor withExponentialBackoff() {
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RetryInterceptor.java
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * Get an instance of {@link RetryInterceptor} that uses full jitter backoff retry strategy.
+     *
+     * @param maxRetries The maximum number of times to retry.
+     * @param baseDelay The delay used as the coefficient for backoffs, also baseDelay will be the first backoff delay.
+     * @param maxDelay The maximum backoff delay before a retry.
+     * @return The retry interceptor.
+     */
+    public static RetryInterceptor exponentialBackoff(int maxRetries, Duration baseDelay, Duration maxDelay) {
```
vcolin7, line 69:
```suggestion
    public static RetryInterceptor withExponentialBackoff(int maxRetries, Duration baseDelay, Duration maxDelay) {
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RetryInterceptor.java
anuchandy: thanks, Victor! `with*` is good with me. 

One thing we might want to consider later is, is it better if we use `createWith*`? two reasons:

1. This is a factory method, so `create` is a common prefix.
2. The prefix `with` is a common prefix used to qualify setters in the builder/fluent pattern e.g. [here](https://github.com/Azure/azure-libraries-for-java/blob/5280b755fd364ec266c50e78c7fd7acfee382547/azure-samples/src/main/java/com/microsoft/azure/management/compute/samples/ConvertVirtualMachineToManagedDisks.java#L47).

but we can revisit it later but what you proposed is better than the current naming.

sdk/core/azure-core/src/main/java/com/azure/android/core/internal/util/ExceptionUtils.java
```diff
@@ -22,6 +22,11 @@
     private ExceptionUtils() {
     }
 
+    /**
+     * A singleton IOException representing cancellation error.
+     */
+    public static final IOException CALL_CANCELLED_IO_EXCEPTION = new IOException("Canceled.");
```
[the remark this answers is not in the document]
anuchandy: I preferred "Cancell****" for our "Type Names" mainly because the same is used in well-known [CancellationToken](https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=netcore-3.1).

Our types are:

1. CancellationToken (public)
2. CancellationTokenImpl (internal)
3. ExceptionUtils.CALL_CANCELLED_IO_EXCEPTION (internal)

The reason we used `Canceled` in the "Exception message" of `IOException("Canceled.")` which we throw onCancel is to consistent with what OkHttp throws from its stack under the same condition, refer [this](https://github.com/square/okhttp/blob/4bfa33e09a4e472f8d8df402507d5bc0c4d3a416/okhttp/src/main/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt#L72), basically, we don't want customers to string match two different strings in case they have any logic based on the error message.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/interceptor/RetryInterceptor.java
vcolin7: I'm also ok with using `createWith`, maybe we can bring this up in our next meeting :)

sdk/core/azure-core/src/main/java/com/azure/android/core/internal/util/ExceptionUtils.java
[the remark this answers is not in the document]
vcolin7: I see, I agree that it makes sense to be consistent with OkHttp. I don't have a strong opinion on using either, however, I think we should have a default for every other case. I'll bring up with the team.

PR #283, opened 2020-08-10
Adding verify link into analyze steps

sima-zhu:


eng/ignore-links.txt
```diff
@@ -0,0 +1,2 @@
+#PlaceHolder for template links
```
weshaggard, line 1: I would remove "not-valid-link" and just leave the comment `# Placeholder file for links that need exceptions from link checking`

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -36,6 +36,11 @@ jobs:
       vmImage: '$(OSVmImage)'
 
     steps:
+
+      - template: ../../../common/pipelines/templates/steps/verify-links.yml
```
weshaggard, line 40: This will only check the links under the service directory. In the other languages there is an "aggregate-reports" pipeline which we used to check links in the other MD files like in the root of the repo. @mitchdenny @bsiegel is there an aggregate reports type pipeline for this repo yet?

sima-zhu:
Will have a follow up for the entire repo lint checker.

sima-zhu:
https://github.com/Azure/azure-sdk/issues/1660

Have an issue to follow up as it is not quick fix here. Will have a new PR if we decided to the way to go.

PR #293, opened 2020-08-18
Created classes for loading configuration values.

vcolin7:


vcolin7:
Will revisit this in the future if we decide to support such functionality for Android.

PR #317, opened 2020-08-25
Added options classes to azure-core and a timeout property to CancellationToken.

### LINKED ISSUE #314: Azure Core Options types
azure-core should contain the following interface hierarchy which will contain global configuration parameters that will be used by all clients.

azure-storage-blob should have Options objects that implement these interfaces - the client builder should accept a StorageBlobClientOptions, and each method that has options should take a per-method Options object (e.g. ListBlobsOptions, DeleteBlobOptions).

```
AzureClientOptions (Base class for client options)
└──StorageBlobClientOptions (Specific client options, passed to builder)

AzureOptions (Base class for client method options)
└──DeleteBlobOptions (Specific non-paginated client method options, passed to the client method)
```

AzureClientOptions should have at least the following members:
```java
    // The API version of the service to invoke.
    String apiVersion;
    // The `ClientLogger` to be used by the service client.
    ClientLogger logger;
    // Options for configuring telemetry sent by the service client.
    TelemetryOptions telemetryOptions;
    // Global transport options.
    TransportOptions transportOptions;
```

TelemetryOptions should have at least the following members:
```java
    // Whether platform information will be omitted from the user agent string sent by the service client.
    boolean telemetryDisabled;
    // An optional user-specified application ID included in the user agent string sent by the service client.
    @Nullable String applicationId;
```

TransportOptions should have at least the following members:
```java
    // Default timeout on any network call. -1 means no timeout. (Should this be a nullable `Double` instead?)
    double timeoutInSeconds;
    // Retry strategy.
    @Nullable RetryStrategy retryStrategy;
```

AzureOptions should have at least the following members:
```java
    // A client-generated, opaque value with 1KB character limit that is recorded in analytics logs.
    // Highly recommended for correlating client-side activities with requests received by the server.
    @Nullable String clientRequestId;
    // A token used to make a best-effort attempt at canceling a request.
    @Nullable CancellationToken cancellationToken;
```

### LINKED ISSUE #315: Use TelemetryOptions in UserAgentInterceptor
Based on the provided TelemetryOptions, configure the UserAgentInterceptor to use the provided application ID, and to include or exclude identifying information. See the implementation in https://github.com/Azure/azure-sdk-for-ios/commit/c45005ff85f27cb6695357d5a0d1bceca83af59d

### LINKED ISSUE #316: Update client libraries to use Options types in Core
After #314, we have a standard hierarchy of Options types that follows our guidelines. We still need to update client libraries to make use of them.

1. Based on the provided TransportOptions, configure the RetryInterceptor to use the provided RetryStrategy.
2. Implement timeouts for CancellationTokens - if the call has not completed in its entirety when the timeout fires, report back to the caller that the service client method call failed because the timeout elapsed. This means you need to check the timer in the pipeline right before the request goes out, and right after it comes back. Even if the request comes back successfully, if the timeout has elapsed, still report back that the request timed out.
3. Based on the provided TransportOptions, use the provided global timeout to attach a CancellationToken with the global timeout value to every request that doesn't have one. If a request does have a user-supplied CancellationToken, but that token lacks a timeout, also update the token with the global timeout value.

vcolin7:
Fixes #314
Fixes #315
Fixes #316


sdk/core/azure-core/src/main/java/com/azure/android/core/http/options/AzureClientOptions.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Options for configuring a {@link ServiceClient}.
+     *
+     * @param apiVersion       The API version of the service to invoke.
+     * @param logger           The {@link ClientLogger} to be used by the {@link ServiceClient}.
+     * @param telemetryOptions Options for configuring telemetry sent by the {@link ServiceClient}.
+     * @param transportOptions Options for configuring calls sent by the {@link ServiceClient}.
+     */
+    public AzureClientOptions(@NonNull String apiVersion,
+                              @Nullable ClientLogger logger,
```
bsiegel, line 28: Should we set this to ClientLogger.NONE (or whatever the equivalent is in Java for the do-nothing logger)? Otherwise I worry there are so many places that we log, that it's too easy to forget to check `logger != null` everywhere.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/options/AzureClientOptions.java
```diff
@@ -0,0 +1,76 @@
...
+                              TelemetryOptions telemetryOptions,
+                              TransportOptions transportOptions) {
+        this.apiVersion = apiVersion;
+        this.logger = logger;
+        this.telemetryOptions = telemetryOptions == null ? new TelemetryOptions(false, null) : telemetryOptions;
```
bsiegel, line 33: Let's have constructor overloads that set these default values rather than specifying them here.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/options/AzureClientOptions.java
vcolin7: We actually have a [default logger](https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/AndroidClientLogger.java) that we can get using [`ClientLogger.getDefault()`](https://github.com/Azure/azure-sdk-for-android/blob/a9c0645976c17c20753d01f8bb3c4dc95c3f3015/sdk/core/azure-core/src/main/java/com/azure/android/core/util/logging/ClientLogger.java#L18). Is it preferred to use the default one or should we provide the option of having a no-op logger as well?

bsiegel: Good question, I'll see what we did in iOS. I would -think- that the default is that logging should be on at WARN level (which should already be the default log level).

vcolin7: It appears that the default level for both iOS and Android is INFO right now.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/options/AzureClientOptions.java
vcolin7: What about the `ClientLogger`? If it's nullable I assume we would have something like this:

* ``` javapublic AzureClientOptions(@NonNull String apiVersion, @Nullable ClientLogger logger)`
* `public AzureClientOptions(@NonNull String apiVersion,
                              @Nullable ClientLogger logger,
                              @NonNull TelemetryOptions telemetryOptions,
                              @NonNull TransportOptions transportOptions)`

sdk/core/azure-core/src/main/java/com/azure/android/core/http/options/AzureClientOptions.java
bsiegel: Ah okay, we will probably want to adjust that at some point.

Checked iOS and we use the default logger if none is provided, and apply the client name as the tag (e.g. "StorageBlobClient").

sdk/core/azure-core/src/main/java/com/azure/android/core/http/options/AzureClientOptions.java
bsiegel: Those look good except I would make ClientLogger @NonNull. Initially we had a default for the logger in Core for iOS as well, but realized that was a bad idea because you need to specify the name of the service client class for tagging log lines. So now each service client MUST define its own default ClientLogger instance. It can still be @Nullable in the concrete client options class (e.g. StorageBlobClientOptions) but the base class in Core should require it to be @NonNull.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/CancellationToken.java
```diff
@@ -25,15 +31,42 @@ public static CancellationToken create() {
...
+     * Reset the {@link CancellationToken} and allow the timeout countdown to be restarted.
+     */
+    public abstract void reset();
+
+    /**
+     * Gets this {@link CancellationToken}'s timeout in seconds.
+     */
+    public Duration getTimeout() {
+        return timeout;
+    }
```
anuchandy, line 63: I see that we want to support a token that triggers a cancellation when timeout happens. It's a "specialized cancellation token" variant of "general cancellation token". 

Why we want to expose the APIs - start|reset -, etc.. specific to the "specialized cancellation token" in the "general cancellation token"? should we not-expose them now and enable it when there is a requirement?

i.e. should we start with a minimal set of public API for the timeout use case? i.e. just 

```java
createWithTimeout(@NonNull Duration timeout);
```

then we hide/remove all other APIs.

thoughts?

anuchandy: Another option to consider is to have a setter ".setCancelTimeout(Duration timeout)" in the CancellationToken.

```java
CancellationToken cancelToken = CancellationToken.create();
token.setCancelTimeout(timeout);
apiCall(params, cancelToken);
```

The "setCancelTimeout" can be called multiple times; it will reset the countdown if token not already canceled. This should help us to cut those APIs, at the same time provide the ability to extend the timeout.

sdk/core/azure-core/src/main/java/com/azure/android/core/internal/util/CancellationTokenImpl.java
```diff
@@ -60,6 +76,38 @@ public void cancel() {
...
     }
 
+    @Override
+    public void start() {
+        if (timeout.equals(Duration.ZERO) || isStarted.get()) {
+            return;
+        }
+
+        UUID expectedRunId = runId;
+        Handler handler = new Handler();
```
anuchandy, line 86: It looks like this Handler is deprecated https://developer.android.com/reference/android/os/Handler#Handler(), we should use the application main looper or executor.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/CancellationToken.java
anuchandy: Also, it is important to note that once a token is canceled, it's canceled forever; we shouldn't implicitly or explicitly allow setting such a token back to be in a non-canceled state.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/options/RequestOptions.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.azure.android.core.http.ServiceClient;
+import com.azure.android.core.util.CancellationToken;
+
+/**
+ * Options for a service operation to be carried out by a {@link ServiceClient}.
+ */
+public class RequestOptions {
+    @Nullable
+    String clientRequestId;
```
anuchandy, line 17: Do we want to expose an option for the user to set the RequestId? In Java SDK, the generation and setting of the request-id header is hidden in the default RequestId policy.

vcolin7:
Closing in favor of creating smaller PRs for solving the issues in the description. Also for separating the `CancellationToken` timeout stuff from the Options classes, since that approach is not definitive yet.

PR #346, opened 2020-09-16
Create a new PR for paging support in Android core

JianpingChen:
Add basic paging support. 

Design goals:
1. Support autorest.android generating Async client. Required classes: Page, AsyncPagedDataRetriever, AsyncPagedDataCollection
2. Support autorest.android generating Sync client. Required classes: Page, PagedDataCollection, PagedDataRetriever, PagedDataResponseCollection, PagedDataResponseRetriever

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
@@ -62,7 +62,7 @@ public class AsyncPagedDataCollection<T, P extends Page<T>> {
                 @Override
                 public void onSuccess(P value, Response response) {
                     if (value == null) {
-                        callback.onSuccess(value, response);
+                        callback.onSuccess(null, response);
                         return;
                     }
 
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
@@ -26,8 +26,8 @@ public class AsyncPagedDataCollection<T, P extends Page<T>> {
     private String firstPageId;
 
     /**
-     * Construction requires an asynchronous paged data provider
-     * @param asyncPagedDataRetriever an asynchronous page data provider
+     * Constructor requires an asynchronous paged data provider
+     * @param asyncPagedDataRetriever an asynchronous paged data provider
      */
     public AsyncPagedDataCollection(@NonNull AsyncPagedDataRetriever<T, P> asyncPagedDataRetriever) {
         Objects.requireNonNull(asyncPagedDataRetriever);
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
@@ -59,7 +59,6 @@ public class AsyncPagedDataCollection<T, P extends Page<T>> {
         Objects.requireNonNull(callback);
         if (firstPageId == null){
             asyncPagedDataRetriever.getFirstPage(new Callback<P>() {
-
                 @Override
                 public void onSuccess(P value, Response response) {
                     if (value == null) {
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
@@ -4,11 +4,14 @@
  */
 package com.azure.android.core.http.responsepaging;
 
+import androidx.annotation.NonNull;
+
 import com.azure.android.core.http.Callback;
 import com.azure.android.core.util.paging.Page;
 
 import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.Objects;
 
 import okhttp3.Response;
 
@@ -18,16 +21,17 @@ import okhttp3.Response;
  * @param <P> Page of items
  */
 public class AsyncPagedDataCollection<T, P extends Page<T>> {
-    private final AsyncPagedDataRetriever<T, P> pagedDataRetriever;
+    private final AsyncPagedDataRetriever<T, P> asyncPagedDataRetriever;
     private LinkedHashMap<String, PageAndResponse<P>> pages = new LinkedHashMap<String, PageAndResponse<P>>();
     private String firstPageId;
 
     /**
      * Construction requires an asynchronous paged data provider
-     * @param pagedDataRetriever an asynchronous page data provider
+     * @param asyncPagedDataRetriever an asynchronous page data provider
      */
-    public AsyncPagedDataCollection(AsyncPagedDataRetriever<T, P> pagedDataRetriever) {
-        this.pagedDataRetriever = pagedDataRetriever;
+    public AsyncPagedDataCollection(@NonNull AsyncPagedDataRetriever<T, P> asyncPagedDataRetriever) {
+        Objects.requireNonNull(asyncPagedDataRetriever);
+        this.asyncPagedDataRetriever = asyncPagedDataRetriever;
     }
 
     private void cacheResponse(P page, Response response) {
@@ -51,9 +55,10 @@ public class AsyncPagedDataCollection<T, P extends Page<T>> {
      * Gets first page in the collection along with the response retrieving the first page
      * @param callback a callback interface for handling the first page and its response
      */
-    public void getFirstPage(Callback<P> callback) {
+    public void getFirstPage(@NonNull Callback<P> callback) {
+        Objects.requireNonNull(callback);
         if (firstPageId == null){
-            pagedDataRetriever.getFirstPage(new Callback<P>() {
+            asyncPagedDataRetriever.getFirstPage(new Callback<P>() {
 
                 @Override
                 public void onSuccess(P value, Response response) {
@@ -84,10 +89,12 @@ public class AsyncPagedDataCollection<T, P extends Page<T>> {
      * @param pageId id of the page
      * @param callback callback interface for handling the page along with its response
      */
-    public void getPage(String pageId, Callback<P> callback) {
+    public void getPage(@NonNull String pageId, @NonNull Callback<P> callback) {
+        Objects.requireNonNull(pageId);
+        Objects.requireNonNull(callback);
         PageAndResponse<P> pageEntry = pages.get(pageId);
         if (pageEntry == null){
-            pagedDataRetriever.getPage(pageId, new Callback<P>() {
+            asyncPagedDataRetriever.getPage(pageId, new Callback<P>() {
                 @Override
                 public void onSuccess(P value, Response response) {
                     if (value == null) {
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
@@ -52,8 +52,8 @@ public class AsyncPagedDataCollection<T, P extends Page<T>> {
     }
 
     /**
-     * Gets first page in the collection along with the response retrieving the first page
-     * @param callback a callback interface for handling the first page and its response
+     * Retrieve the first page of the collection along with the response it comes with
+     * @param callback callback interface for handling the first page and the response it comes with
      */
     public void getFirstPage(@NonNull Callback<P> callback) {
         Objects.requireNonNull(callback);
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
@@ -16,9 +16,9 @@ import java.util.Objects;
 import okhttp3.Response;
 
 /**
- * Represents a collection of pages where page data and response for retrieving it can be returned asynchronously
+ * Represents a collection of pages where page data and the response it comes with it can be obtained asynchronously
  * @param <T> type of the items in the page
- * @param <P> Page of items
+ * @param <P> type of the page
  */
 public class AsyncPagedDataCollection<T, P extends Page<T>> {
     private final AsyncPagedDataRetriever<T, P> asyncPagedDataRetriever;
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
@@ -1,0 +1,121 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+package com.azure.android.core.http.responsepaging;
+
+import com.azure.android.core.http.Callback;
+import com.azure.android.core.util.paging.Page;
+
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+
+import okhttp3.Response;
+
+/**
+ * Represents a collection of pages where page data and response for retrieving it can be returned asynchronously
+ * @param <T> type of the items in the page
+ * @param <P> Page of items
+ */
+public class AsyncPagedDataCollection<T, P extends Page<T>> {
+    private final AsyncPagedDataRetriever<T, P> pagedDataRetriever;
+    private LinkedHashMap<String, PageAndResponse<P>> pages = new LinkedHashMap<String, PageAndResponse<P>>();
+    private String firstPageId;
+
+    /**
+     * Construction requires an asynchronous paged data provider
+     * @param pagedDataRetriever an asynchronous page data provider
+     */
+    public AsyncPagedDataCollection(AsyncPagedDataRetriever<T, P> pagedDataRetriever) {
+        this.pagedDataRetriever = pagedDataRetriever;
+    }
+
+    private void cacheResponse(P page, Response response) {
+        if (page.getPageId() == null) {
+            return;
+        }
+        pages.put(page.getPageId(), new PageAndResponse<P>(page, response));
+
+        // setting previous page id should simplify implementation for androidx.arch.DataSource
+        final Iterator<PageAndResponse<P>> iterator = pages.values().iterator();
+        while(iterator.hasNext()){
+            final P existingPage = iterator.next().page;
+            if (page.getPageId().equals(existingPage.getNextPageId())){
+                page.setPreviousPageId(existingPage.getPageId());
+                break;
+            }
+        }
+    }
+
+    /**
+     * Gets first page in the collection along with the response retrieving the first page
+     * @param callback a callback interface for handling the first page and its response
+     */
+    public void getFirstPage(Callback<P> callback) {
+        if (firstPageId == null){
+            pagedDataRetriever.getFirstPage(new Callback<P>() {
+
+                @Override
+                public void onSuccess(P value, Response response) {
+                    if (value == null) {
+                        callback.onSuccess(value, response);
+                        return;
+                    }
+
+                    firstPageId = value.getPageId();
+                    cacheResponse(value, response);
+                    callback.onSuccess(value, response);
+                }
+
+                @Override
+                public void onFailure(Throwable t, Response response) {
+                    callback.onFailure(t, response);
+                }
+            });
+        }
+        else {
+            PageAndResponse<P> firstPageResponse = pages.get(firstPageId);
+            callback.onSuccess(firstPageResponse.page, firstPageResponse.response);
+        }
+    }
+
+    /**
+     * Retrieve a page by its id
+     * @param pageId id of the page
+     * @param callback callback interface for handling the page along with its response
+     */
+    public void getPage(String pageId, Callback<P> callback) {
+        PageAndResponse<P> pageEntry = pages.get(pageId);
+        if (pageEntry == null){
+            pagedDataRetriever.getPage(pageId, new Callback<P>() {
+                @Override
+                public void onSuccess(P value, Response response) {
+                    if (value == null) {
+                        callback.onSuccess(null, response);
+                    }
+
+                    cacheResponse(value, response);
+                    callback.onSuccess(value, response);
+                }
+
+                @Override
+                public void onFailure(Throwable t, Response response) {
+                    callback.onFailure(t, response);
+                }
+            });
+        }
+        else {
+            callback.onSuccess(pageEntry.page, pageEntry.response);
+        }
+    }
+
+    private static class PageAndResponse<P> {
+        public final P page;
+        public final Response response;
+
+        public PageAndResponse(P page, Response response) {
+            this.page = page;
+            this.response = response;
+        }
+    }
+}
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
@@ -84,9 +84,9 @@ public class AsyncPagedDataCollection<T, P extends Page<T>> {
     }
 
     /**
-     * Retrieve a page by its id
+     * Retrieve a page with the given id
      * @param pageId id of the page
-     * @param callback callback interface for handling the page along with its response
+     * @param callback callback interface for handling the page along with the response it comes with
      */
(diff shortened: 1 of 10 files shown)


sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
```diff
@@ -0,0 +1,121 @@
...
+                    callback.onFailure(t, response);
+                }
+            });
+        }
+        else {
+            callback.onSuccess(pageEntry.page, pageEntry.response);
+        }
+    }
+
+    private static class PageAndResponse<P> {
```
bsiegel, line 112: We should be using `com.azure.android.core.http.Response<T>` as the type to represent a response with an associated strongly-typed response body object. Is there any reason we aren't doing that here?

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataRetriever.java
```diff
@@ -0,0 +1,51 @@
...
+     * @param callback callback function to handle the page data along with the response for retrieving it
+     */
+    public abstract void getFirstPage(Callback<P> callback);
+
+    /**
+     * Retrieve a page with its id
+     * @param pageId id of the page
+     * @param callback callback function to handle the page data along with the response for retrieving it
+     */
+    public abstract void getPage(String pageId, Callback<P> callback);
```
bsiegel, line 50: How is the pageId supposed to be communicated to the developer? Is it an opaque identifier (e.g. GUID)? I would expect `getPage(int pageNumber, Callback callback)` or something similar, otherwise the developer has to do book-keeping.

Alternatively, why not keep the current page number in the instance's local state, and have a `getNextPage(Callback callback)` method (and a simple getter, `getCurrentPageNumber()`or some such) instead?

bsiegel: I guess this comment isn't specific to this file, but just to the general design.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseCollection.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Represents a collection of pages where page and the response for retrieving it can be returned synchronously
+ * @param <T> type of the items contained in the page
+ * @param <P> page that contains the items
+ */
+public class PagedDataResponseCollection<T, P extends Page<T>> {
```
bsiegel, line 19: Is this the sync version of `AsyncPagedDataCollection`? If so, they should have similar names. E.g. `PagedDataCollection`.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseRetriever.java
```diff
@@ -0,0 +1,47 @@
...
+import com.azure.android.core.util.paging.Page;
+
+import java.util.Objects;
+
+/**
+ * Represents a synchronized paged data provider with fixed page size that returns both page and response
+ * @param <T> type of the items in the page
+ * @param <P> type of the page
+ */
+public abstract class PagedDataResponseRetriever<T, P extends Page<T>> {
```
bsiegel, line 17: Likewise if this is the sync version of `AsyncPagedDataRetriever`, let's make this `PagedDataRetriever`.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseRetriever.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package com.azure.android.core.http.responsepaging;
+
+import com.azure.android.core.http.Response;
+import com.azure.android.core.util.paging.Page;
+
+import java.util.Objects;
+
+/**
+ * Represents a synchronized paged data provider with fixed page size that returns both page and response
```
bsiegel, line 13:
```suggestion
 * Represents a synchronous paged data provider with fixed page size that returns both page and response
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
JianpingChen: Sorry Brandon, This branch is messed up somehow. Please put comment in https://github.com/Azure/azure-sdk-for-android/pull/317 instead. I'll still read here but more people may have been directed there.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/paging/PagedDataCollection.java
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+package com.azure.android.core.util.paging;
+
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+
+/**
+ * Represents a collection of paged data with fixed page size where pages can be retrieved synchrounously
```
bsiegel, line 11:
```suggestion
 * Represents a collection of paged data with fixed page size where pages can be retrieved synchronously
```

sdk/core/azure-core/src/main/java/com/azure/android/core/util/paging/PagedDataCollection.java
```diff
@@ -0,0 +1,66 @@
...
+ * @param <T> type of the items contained in the page
+ * @param <P> page that contains the items
+ */
+public class PagedDataCollection<T, P extends Page<T>> {
```
bsiegel, line 15: Not clear why there is a separate set of classes for paged data vs. paged data + response, but only for sync (for async there is only paged data + response).

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
JianpingChen: My bad. I thought this was the old one. ;-)

The reason for PageAndResponse is this collection has an interface where getFirstPage and getPage both take a Callback<P> where both Page and okhttp3.Response are supposed to be used. Unless we change the callback signature, we'd have to keep both in order to hand back the okhttp3.Response.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataRetriever.java
JianpingChen: For the need of autorest.android, the String id is meant to be opaque. I even had a page iterator function in the collection trying to hide the pageId altogether but it turned out to be overly complicated (in thread synchronization). So we agreed to remove the iterator implementation. 

We looked at using this interface to implement androidx.arch.DataSource in my initial Draft PR. Having an opaque page id is more convenient than int id for that particular purpose. That's the reason we settled with an opaque string id.

c.f https://developer.android.com/reference/android/arch/paging/PageKeyedDataSource#loadBefore(android.arch.paging.PageKeyedDataSource.LoadParams%3CKey%3E,%20android.arch.paging.PageKeyedDataSource.LoadCallback%3CKey,%20Value%3E)

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseCollection.java
JianpingChen: There is a PagedDataCollection in util.paging. We can rename AsyncPagedDataCollection to AsyncPagedDataResponseCollection

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseRetriever.java
JianpingChen: There is a PagedDataRetriever in util.paging which does not return Response.

sdk/core/azure-core/src/main/java/com/azure/android/core/util/paging/PagedDataCollection.java
JianpingChen: When autorest.android generates sync client, for each function in underlying service, e.g. service.getBlah(), sync client is supposed to have two versions of the same function in its interface, syncClient.getBlah(): Blah, and syncClient.getBlahWithResponse(): Response\<Blah\>. That's why we have two sets for sync client.

JianpingChen: For async client, the Callback\<Blah\> can return both Blah and okhttp3.Response. Therefore there is no need to have two function signatures for each service.getBlah function.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataRetriever.java
anuchandy: Thinking a bit more about it: 

The paged collection = retrieve pages from service via network + caching of the pages.

Wondering, whether it makes sense to have the following signature:

```java
// group_1: calls go over network.
void getFirstPage(Callback<Page<T>> callback);
void getNextPage(Callback<Page<T>> callback);
```

```java
// group_2: an immutable map of cached pages.
Map<String, Page<T>> getCachedPages();
// an immutable list of items in the cached pages.
List<T> getCachedItems();
```

Only the APIs that belongs to the group_1 provides access to the Response object.

bsiegel:
@JianpingChen was there anything else we needed to do to this PR post today's discussion?

sdk/core/azure-core/src/test/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollectionTest.java
```diff
@@ -0,0 +1,131 @@
+package com.azure.android.core.http.responsepaging;
```
vcolin7, line 1: We need to add the license header here.

sdk/core/azure-core/src/test/java/com/azure/android/core/http/responsepaging/PagedDataResponseCollectionTest.java
```diff
@@ -0,0 +1,114 @@
+package com.azure.android.core.http.responsepaging;
```
vcolin7, line 1: We need to add the license header here.

sdk/core/azure-core/src/test/java/com/azure/android/core/util/paging/PagedDataCollectionTest.java
```diff
@@ -0,0 +1,109 @@
+package com.azure.android.core.util.paging;
```
vcolin7, line 1: We need to add the license header here.

JianpingChen:
> @JianpingChen was there anything else we needed to do to this PR post today's discussion?

@bsiegel As far as I remember, no. The idea of making it Page<K, T> was rejected. Arguments around setters for nexpPage and previousPage seems to settle with "as is", unless I misunderstood anything. We can make it clear all collections are caching in JavaDoc if that is required.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
bsiegel: Leaving as-is for this preview.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataRetriever.java
bsiegel: Leaving as-is for this preview.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseCollection.java
bsiegel: Okay, I understand why they are the way they are, no need to rename.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseRetriever.java
bsiegel: Leaving as is

sdk/core/azure-core/src/main/java/com/azure/android/core/util/paging/PagedDataCollection.java
bsiegel: Leaving as-is for this preview.

bsiegel:
Sounds good, can you resolve the remaining comments (just typos and missing headers)? Then we can go ahead and merge.

anuchandy:
Update: This feature is pushed to post preview_1 as per offline discussions within the crew.

JianpingChen:
@bsiegel @vcolin7 @anuchandy I resolved the left over comments. Please take another look and approve if no further concerns.

vcolin7:
@JianpingChen You probably don't have write access to the public repo. I'm not sure who's in charge of assigning such permissions. I will review later today :)

weshaggard:
@JianpingChen @RezaJooyandeh can grant you the permission to be able to merge.

RezaJooyandeh:
@JianpingChen Now do have access :)

sdk/core/azure-core/src/main/java/com/azure/android/core/util/paging/PagedDataCollection.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param pagedDataRetriever paged data provider with fixed page size
+     */
+    public PagedDataCollection(@NonNull PagedDataRetriever<T, P> pagedDataRetriever) {
+        Objects.requireNonNull(pagedDataRetriever);
+        this.pagedDataRetriever = pagedDataRetriever;
+    }
+
+    /**
+     * Retrieve the first page in the collection
+     * @return First page of the collection
```
vcolin7, line 34:
```suggestion
     * Retrieve the first page of the collection
     * @return the first page of the collection
```

sdk/core/azure-core/src/main/java/com/azure/android/core/util/paging/PagedDataCollection.java
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public P getFirstPage() {
+        if (firstPageId != null) {
+            return pages.get(firstPageId);
+        }
+        P firstPage = pagedDataRetriever.getFirstPage();
+        firstPageId = firstPage.getPageId();
+        pages.put(firstPage.getPageId(), firstPage);
```
vcolin7, line 42: nit; We can use the value we just set.
```suggestion
        pages.put(firstPageId, firstPage);
```

sdk/core/azure-core/src/main/java/com/azure/android/core/util/paging/PagedDataCollection.java
```diff
@@ -0,0 +1,70 @@
...
+        return  firstPage;
+    }
+
+    /**
+     * Retrieve a page with its id
+     * @param pageId id of the page
+     * @return page of data with the requested id
+     */
+    public P getPage(@NonNull String pageId) {
+        P page = pages.get(pageId);
```
vcolin7, line 52: What happens if the given `pageId` is `null`? The annotation will not stop users from passing such a value.
```suggestion
        Objects.requireNonNull(pageId);
        P page = pages.get(pageId);
```

sdk/core/azure-core/src/main/java/com/azure/android/core/util/paging/PagedDataCollection.java
```diff
@@ -0,0 +1,70 @@
...
+        // setting previous page id should simplify implementation for androidx.arch.DataSource
+        final Iterator<P> iterator = pages.values().iterator();
+        while(iterator.hasNext()){
+            final P existingPage = iterator.next();
+            if (pageId.equals(existingPage.getNextPageId())) {
+                page.setPreviousPageId(existingPage.getPageId());
+                break;
+            }
+        }
+        pages.put(page.getPageId(), page);
```
vcolin7, line 67: nit; we can just use the `pageId` given as an input parameter.
```suggestion
        pages.put(pageId, page);
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseCollection.java
```diff
@@ -0,0 +1,83 @@
...
+            return pages.get(firstPageId);
+        }
+        Response<P> firstPageResponse = pagedDataRetriever.getFirstPage();
+        if (firstPageResponse != null
+            && firstPageResponse.getValue() != null
+            && firstPageResponse.getValue().getPageId() != null) {
+            firstPageId = firstPageResponse.getValue().getPageId();
+            pages.put(firstPageId, firstPageResponse);
+        }
+        return  firstPageResponse;
```
vcolin7, line 51: nit;
```suggestion
        return firstPageResponse;
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseCollection.java
```diff
@@ -0,0 +1,83 @@
...
+            // setting previous page id should simplify implementation for androidx.arch.DataSource
+            final Iterator<Response<P>> iterator = pages.values().iterator();
+            while (iterator.hasNext()) {
+                final P existingPage = iterator.next().getValue();
+                if (pageId.equals(existingPage.getNextPageId())) {
+                    pageResponse.getValue().setPreviousPageId(existingPage.getPageId());
+                    break;
+                }
+            }
+            pages.put(pageResponse.getValue().getPageId(), pageResponse);
```
vcolin7, line 79: nit; We can use the `pageId` given as an input parameter.
```suggestion
            pages.put(pageId, pageResponse);
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataRetriever.java
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Retrieve the first page in the collection
+     * @param callback callback function to handle the page data along with the response for retrieving it
+     */
+    public abstract void getFirstPage(Callback<P> callback);
+
+    /**
+     * Retrieve a page with its id
+     * @param pageId id of the page
+     * @param callback callback function to handle the page data along with the response for retrieving it
```
vcolin7, line 26:
```suggestion
     * Retrieve a page with the given id
     * @param pageId id of the page
     * @param callback callback function to handle the page data along with the response it comes with
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Objects;
+
+import okhttp3.Response;
+
+/**
+ * Represents a collection of pages where page data and response for retrieving it can be returned asynchronously
+ * @param <T> type of the items in the page
+ * @param <P> Page of items
```
vcolin7, line 21:
```suggestion
 * Represents a collection of pages where page data and the response it comes with it can be obtained asynchronously
 * @param <T> type of the items in the page
 * @param <P> type of the page
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
```diff
@@ -0,0 +1,127 @@
...
+ */
+public class AsyncPagedDataCollection<T, P extends Page<T>> {
+    private final AsyncPagedDataRetriever<T, P> asyncPagedDataRetriever;
+    private LinkedHashMap<String, PageAndResponse<P>> pages = new LinkedHashMap<String, PageAndResponse<P>>();
+    private String firstPageId;
+
+    /**
+     * Construction requires an asynchronous paged data provider
+     * @param asyncPagedDataRetriever an asynchronous page data provider
```
vcolin7, line 30:
```suggestion
     * Constructor requires an asynchronous paged data provider
     * @param asyncPagedDataRetriever an asynchronous paged data provider
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
```diff
@@ -0,0 +1,127 @@
...
+            if (page.getPageId().equals(existingPage.getNextPageId())){
+                page.setPreviousPageId(existingPage.getPageId());
+                break;
+            }
+        }
+    }
+
+    /**
+     * Gets first page in the collection along with the response retrieving the first page
+     * @param callback a callback interface for handling the first page and its response
```
vcolin7, line 56:
```suggestion
     * Retrieve the first page of the collection along with the response it comes with
     * @param callback callback interface for handling the first page and the response it comes with
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseCollection.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public Response<P> getFirstPage() {
+        if (firstPageId != null) {
+            return pages.get(firstPageId);
+        }
+        Response<P> firstPageResponse = pagedDataRetriever.getFirstPage();
+        if (firstPageResponse != null
+            && firstPageResponse.getValue() != null
+            && firstPageResponse.getValue().getPageId() != null) {
+            firstPageId = firstPageResponse.getValue().getPageId();
```
vcolin7, line 48: How likely is it that the `pageId` turns out to be `null`? We don't check for nullness in the other PagedCollection classes.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
```diff
@@ -0,0 +1,127 @@
...
+        else {
+            PageAndResponse<P> firstPageResponse = pages.get(firstPageId);
+            callback.onSuccess(firstPageResponse.page, firstPageResponse.response);
+        }
+    }
+
+    /**
+     * Retrieve a page by its id
+     * @param pageId id of the page
+     * @param callback callback interface for handling the page along with its response
```
vcolin7, line 89:
```suggestion
     * Retrieve a page with the given id
     * @param pageId id of the page
     * @param callback callback interface for handling the page along with the response it comes with
```

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
vcolin7: Adding to this for later discussion: We can definitely handle the given `okhttp3.Response` in the callback and map its contents to our own `Response<T>` type. I think we did that in Storage at some point.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/AsyncPagedDataCollection.java
```diff
@@ -0,0 +1,127 @@
...
+     * @param callback a callback interface for handling the first page and its response
+     */
+    public void getFirstPage(@NonNull Callback<P> callback) {
+        Objects.requireNonNull(callback);
+        if (firstPageId == null){
+            asyncPagedDataRetriever.getFirstPage(new Callback<P>() {
+                @Override
+                public void onSuccess(P value, Response response) {
+                    if (value == null) {
+                        callback.onSuccess(value, response);
```
vcolin7, line 65: For consistency with `getPage()` below.
```suggestion
                        callback.onSuccess(null, response);
```

vcolin7 (commented):
Logic-wise most of the code looks great. I added a couple comments about checking for nullness and a lot of small observations about the Javadoc.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/responsepaging/PagedDataResponseCollection.java
JianpingChen: Added null check wherever missing. Hope this resolves it

vcolin7 (approved):
Aside from some Javadoc details it looks good to me :)

PR #348, opened 2020-09-17
Replaced relative link with absolute links and remove locale

sima-zhu:
The PR is to turn on the link check flag.

1. Disable the relative links.
1. Remove locale "en-us" from link.

Follow up issue: https://github.com/Azure/azure-sdk/issues/1730

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -40,6 +40,7 @@ jobs:
       - template: ../../../common/pipelines/templates/steps/verify-links.yml
         parameters:
           Directory: ''
+          CheckLinkGuidance: $true
 
       - script: |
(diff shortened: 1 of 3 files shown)


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -40,6 +40,7 @@ jobs:
       - template: ../../../common/pipelines/templates/steps/verify-links.yml
         parameters:
           Directory: ''
+          CheckLinkGuidance: $true
```
weshaggard, line 43: Can we also add this to the aggregate reports pipeline and queue that build to verify all the links in the repo?

weshaggard: Nevermind this is the android repo which doesn't have that check.

weshaggard (approved):

weshaggard:
Link check passed so merging.

PR #351, opened 2020-09-18
Fix the group ID (missing from scripts).

mitchdenny:
Fixes the group Id for late state release pipeline.


bsiegel (approved):
LGTM :shipit:

PR #383, opened 2020-10-06
Update readmes for release

bsiegel:
This should be merged after chat is published to maven central.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,10 +19,15 @@ For your convenience, each service has a separate set of libraries that you can 
 Currently, the client libraries are in **beta**. These libraries follow the [Azure SDK Design Guidelines for Android](https://azure.github.io/azure-sdk/android_introduction.html) and share a number of core features such as HTTP retries, logging, transport protocols, authentication protocols, etc., so that once you learn how to use these features in one client library, you will know how to use them in other client libraries. You can learn about these shared features in [azure-core](https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/core/azure-core/README.md).
 
 The following libraries are currently in **beta**:
-- [azure-communication-common](https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/communication/azure-communication-common): 1.0.0-beta.1
+
+#### Core
 - [azure-core](https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/core/azure-core): 1.0.0-beta.2
 
-> Note: The SDK is currently in **beta**. The API surface and feature sets are subject to change at any time before **GA**. We do not currently recommend them for production use.
+#### Azure Communication Services
+- [azure-communication-chat](https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/communication/azure-communication-chat): 1.0.0-beta.2
+- [azure-communication-common](https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/communication/azure-communication-common): 1.0.0-beta.1
+
+> Note: The SDK is currently in **beta**. The API surface and feature sets are subject to change at any time before they become generally available. We do not currently recommend them for production use.
 
 ### Install the libraries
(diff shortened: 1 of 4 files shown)


sdk/core/azure-core/README.md
```diff
@@ -22,14 +22,16 @@ as an example of how to use the shared azure-core components in your client libr
   APIs that would require the Java 8+ API desugaring provided by Android Gradle plugin 4.0.0.
 * You must have an [Azure subscription](https://azure.microsoft.com/free/) to use this library.
 
+### Versions available
```
anuchandy, line 25: The H3 `Versions available` says "Versions", but we point only one, i.e. the latest version

README.md
```diff
@@ -19,10 +19,15 @@ For your convenience, each service has a separate set of libraries that you can
...
+
+#### Core
 - [azure-core](https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/core/azure-core): 1.0.0-beta.2
 
-> Note: The SDK is currently in **beta**. The API surface and feature sets are subject to change at any time before **GA**. We do not currently recommend them for production use.
+#### Azure Communication Services
+- [azure-communication-chat](https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/communication/azure-communication-chat): 1.0.0-beta.2
+- [azure-communication-common](https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/communication/azure-communication-common): 1.0.0-beta.1
+
+> Note: The SDK is currently in **beta**. The API surface and feature sets are subject to change at any time before they become generally available. We do not currently recommend them for production use.
```
anuchandy, line 30: so the idea is to have the root README list all available SDKs, which can grow? a bit different from java but not against it.

sdk/communication/azure-communication-chat/README.md
```diff
@@ -1,20 +1,23 @@
...
+| [API reference documentation](https://azure.github.io/azure-sdk-for-android/sdk/communication/azure-communication-chat/azure-communication-chat/index.html)
+| [Product documentation](https://docs.microsoft.com/azure/communication-services/overview)
 
 # Getting started
 
 ## Prerequisites
 
-* You must have an [Azure subscription](https://azure.microsoft.com/free/) to use this library.
-* An Azure Communication Resource, learn how to create one from [Create an Azure Communication Resource](https://docs.microsoft.com/azure/communication-services/quickstarts/create-communication-resource).
+* You must have an [Azure subscription](https://azure.microsoft.com/free/) and a [Communication Services resource](https://docs.microsoft.com/azure/communication-services/quickstarts/create-communication-resource) to use this library.
```
anuchandy, line 13: just curious, why are we removing the need for azure communication resource from prerequisite?

README.md
bsiegel: Right now yes, eventually that should be moved to here: https://azure.github.io/azure-sdk/releases/latest/

sdk/communication/azure-communication-chat/README.md
anuchandy: never mind, I see it's just reformatted.

bsiegel: It's just combined into one sentence to match the azure-communication-common readme.

anuchandy (approved):

sdk/core/azure-core/README.md
bsiegel: That is a little confusing, agreed. I believe the template requires a "Versions available" header, but we don't have an easy way to link to a filtered list of all releases of just this library. We could do what I did in iOS and just say. "Older versions are also available." with a link to the GH release page. What do you think?

anuchandy: ah, was not aware of template requirement. I'm good at following the same iOS did.

vcolin7 (approved):
LGTM :)

PR #389, opened 2020-10-07
Added support to get blob tags

gapra-msft:
Merge after set tier


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
```diff
@@ -701,6 +705,90 @@ void delete(String containerName,
...
+            null,
+            timeout,
+            version,
+            null,
+            requestId,
+            cancellationToken,
+            new CallbackWithHeader<BlobTags, BlobGetTagsHeaders>() {
+                @Override
+                public void onSuccess(BlobTags result, BlobGetTagsHeaders header, Response response) {
+                    Map<String, String> tags = null;
```
gapra-msft, line 775: Should we return empty map or null here if there are no tags?

gapra-msft: We should probably bump the XMS_VERSION to the 73 version since gettags was not supported in 2019-02-02 I dont think

rickle-msft: I would think an empty map, but there might be a guideline for this

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -508,6 +512,73 @@ BlobDeleteResponse deleteWithRestResponse(String containerName,
             cancellationToken);
     }
 
+    /**
+     * Gets tags associated with a blob.
+
```
rickle-msft, line 517: Extra new line

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -508,6 +512,73 @@ BlobDeleteResponse deleteWithRestResponse(String containerName,
...
+     * @param containerName     The container name.
+     * @param blobName          The blob name.
+     * @return The blob's tags.
+     */
+    public Map<String, String> getBlobTags(String containerName,
+                                           String blobName) {
+        BlobTags response = this.storageBlobServiceClient.getTags(containerName,
+            blobName);
+        Map<String, String> tags = null;
+        if (response.getBlobTagSet() != null) {
```
rickle-msft, line 527: Is there any way we can push this conversion into the intern method?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/models/BlobGetTagsHeaders.java
```diff
@@ -0,0 +1,4 @@
+package com.azure.android.storage.blob.models;
+
+public class BlobGetTagsHeaders {
+}
```
rickle-msft, line 4: Why is this empty?

rickle-msft (changes requested):
Fill in the header file and everything else looks good!

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/models/BlobGetTagsHeaders.java
gapra-msft: @rickle-msft Filled it in, Can you take another look?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
gapra-msft: It looks like tagset will never be null from the generated code so I think this is fine?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
gapra-msft: I hesitate to do that cause if/when there is a generator for Android it will probably replace everything in the Impl file, so I want to keep it as "generated" as possible

gapra-msft: Discussed offline, going to move this into a ModelHelper class

rickle-msft (approved):
Live getTags tests coming with the setTags PR

PR #393, opened 2020-10-07
Set tags

rickle-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -1115,6 +1115,20 @@ private BlobSetTagsResponse setBlobTagsWithRestResponseIntern(String containerNa
         BlobTags blobTags = new BlobTags();
         blobTags.setBlobTagSet(blobTagSet);
 
+        RequestBody tagsBody;
+        try {
+            tagsBody = RequestBody.create(MediaType.get("application/xml; charset=utf-8"),
+                serializerAdapter.serialize(blobTags, SerializerFormat.XML));
```
anuchandy, line 1121: yes, this is needed for android :) a bit different from Java RestProxy.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -1013,6 +1093,116 @@ public void onFailure(@NonNull Call<Void> call, @NonNull Throwable t) {
...
+        ((CancellationTokenImpl) cancellationToken).registerOnCancel(() -> {
+            call.cancel();
+        });
+
+        if (callback != null) {
+            executeCall(call, new retrofit2.Callback<ResponseBody>() {
+                @Override
+                public void onResponse(@NonNull Call<ResponseBody> call, @NonNull Response<ResponseBody> response) {
+                    if (response.isSuccessful()) {
+                        if (response.code() == 200) {
```
anuchandy, line 1154: Seems as per swagger, valid response code is [204](https://github.com/Azure/azure-sdk-for-java/blob/9de1e77a1329ad9291e261f737dd7db00209d6d6/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/BlobsImpl.java#L219)

rickle-msft: Nice catch :)

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
```diff
@@ -236,6 +249,140 @@ public void cancel(Context context, long transferId) {
...
     }
 
+    /**
+     * Creates a new container within a storage account. If a container with the same name already exists, the operation
+     * fails.
+     *
+     * @param containerName The container name.
+     * @param callback      Callback that receives the response.
+     */
+    public void createContainer(String containerName,
```
vcolin7, line 259: I don't know why the PR includes all a ton of other things, I guess I should just ignore them? I assume this PR only introduces the `setTags` operations, correct?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -59,20 +77,224 @@
...
+import retrofit2.http.Tag;
 
 /**
  * PACKAGE PRIVATE CLASS AND METHODS
  */
 final class StorageBlobServiceImpl {
     private final StorageBlobService service;
     private final SerializerAdapter serializerAdapter;
-    private static String XMS_VERSION = "2019-02-02";
+    private static String XMS_VERSION = "2019-12-12";
```
vcolin7, line 88: Is this the newest Storage version? Where do we check this? Actually, I see different versions than what we had [here](https://github.com/Azure/azure-rest-api-specs/tree/master/specification/storage/resource-manager/Microsoft.Storage/stable).

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -1013,6 +1660,119 @@ public void onFailure(@NonNull Call<Void> call, @NonNull Throwable t) {
...
+            } else {
+                throw new RuntimeException(ioe);
+            }
+        }
+
+        Call<ResponseBody> call = service.setBlobTags(containerName,
+            blobName,
+            timeout,
+            versionId,
+            null, // TODO: calculate Md5?
```
vcolin7, line 1706: We probably need to consider this in [the MD5 PR](https://github.com/Azure/azure-sdk-for-android/pull/414).

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
rickle-msft: correct

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
rickle-msft: The link you shared is for the resource manager/management plane. The version update here is for the data plane. It's not quite the latest, but it's the version that enabled tags.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
rickle-msft: correct

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
rickle-msft: On second thought, this isn't actually possible for us at this point because it's the md5 of the payload, but we can't calculate that until after the body is serialized, at which point it's out of our control. We could consider adding an interceptor that does that, but it would be a separate PR I think.

vcolin7 (approved):

PR #414, opened 2020-10-20
Added support to automatically compute Md5

gapra-msft:
I think the best way to test the transfer client computeMd5 functionality might be an instrumentation test.


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -1839,6 +1848,18 @@ private BlockBlobsStageBlockResponse stageBlockWithRestResponseIntern(String con
...
         final String comp = "block";
+
+        if (computeMd5) {
+            if (transactionalContentMD5 != null) {
+                throw new IllegalArgumentException("'transactionalContentMD5' can not be set when 'computeMd5' is true.");
+            }
+            try {
+                transactionalContentMD5 = MessageDigest.getInstance("MD5").digest(blockContent);
+            } catch (NoSuchAlgorithmException e) {
+                throw new RuntimeException(e);
```
vcolin7, line 1859: I wonder why we would prefer to throw this as a `RuntimeException` instead of the original type.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/UploadRequest.java
```diff
@@ -63,14 +67,23 @@ String getContainerName() {
...
      */
     String getBlobName() {
         return this.blobName;
     }
 
+    /**
+     * Get the name of the target blob holding uploaded file.
+     *
+     * @return The blob name.
+     */
```
vcolin7, line 82: These two Javadoc entries should be swapped.

vcolin7 (changes requested):
Just added one question and requested to correct the Javadoc in one file. Besides that this looks good to me :)

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/transfer/UploadRequest.java
gapra-msft: Whoops thanks for catching that

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
gapra-msft: This is what we do in Java usually. I think its just a remnant of some logging deficiency that used to exist and Im still not used to keeping the original type. Let me change to just throw the original exception.

gapra-msft: Hmm if I change it to just throw its mad that Im not declaring the exception as part of the method type.

vcolin7: Ok, let's leave it like this for now :)

vcolin7 (approved):

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobServiceImpl.java
```diff
@@ -2197,6 +2205,7 @@ private BlockBlobsStageBlockResponse stageBlockWithRestResponseIntern(String con
                                                                           byte[] blockContent,
                                                                           byte[] transactionalContentMD5,
                                                                           byte[] transactionalContentCrc64,
+                                                                          boolean computeMd5,
```
rickle-msft, line 2208: I wonder if this should be a Boolean instead of boolean?

gapra-msft: True, I think Boolean is easier for people

gapra-msft: Fixed

rickle-msft (approved):

PR #424, opened 2020-10-23
Added ability to set service version on builder and added the request id interceptor instead of passing through the API

gapra-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/BlobServiceVersion.java
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.storage.blob;
+
+
+/**
+ * The versions of Azure Storage Blob supported by this client library.
+ */
+public enum BlobServiceVersion {
```
gapra-msft, line 10: Will we have a core ServiceVersion type this can inherit from like in Java?

vcolin7: I think that will be the case. What do you think @anuchandy, @bsiegel?

vcolin7 (approved):
LGTM!

rickle-msft (approved):

PR #438, opened 2020-10-29
Added support for conditional tags

gapra-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
```diff
@@ -978,14 +974,15 @@ public void getBlobTags(String containerName,
                             String blobName,
                             String snapshot,
                             Integer timeout,
+                            String ifTags,
```
vcolin7, line 977: Is `ifTags` the same as `tagsConditions`? If so, why the name change?

vcolin7 (commented):
I'm not entirely sure what conditional tags are, can you give me a hand? :)

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
gapra-msft: Whoops that is meant to be tagsConditions. Yeah they are one and the same, but ifTags doesn't really give a user a good understanding of what it is (ifTags is a SQL like statement that allows a user to execute a request if the blob has the appropriate tags). We had a discussion with the arch board when this feature was released, and tagsConditions was decided as a better name.

gapra-msft: As for what tags is, it's kind of like a metadata value you can assign to multiple resources to group them

gapra-msft: https://docs.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations#tags-conditional-operations

vcolin7 (approved):

rickle-msft (approved):

PR #441, opened 2020-10-30
Added async tests for blob

gapra-msft:


sdk/storage/azure-storage-blob/src/test/java/com/azure/android/storage/blob/BlobTest.java
```diff
@@ -288,6 +290,32 @@ public void getPropertiesError() {
...
+    public void getPropertiesAsync() {
+        // Setup
+        CountDownLatch latch = new CountDownLatch(1);
+
+        // When
+        asyncClient.getBlobProperties(containerName, blobName, new CallbackWithHeader<Void, BlobGetPropertiesHeaders>() {
+                @Override
+                public void onSuccess(Void result, BlobGetPropertiesHeaders header, Response response) {
+                    assertEquals(200, response.code());
+                    latch.countDown();
```
vcolin7, line 303: Should we check that the header's contents are correct like we check the body contents in `rawRownloadAsync()`?

sdk/storage/azure-storage-blob/src/test/java/com/azure/android/storage/blob/BlobTest.java
```diff
@@ -929,4 +1042,30 @@ public void onFailure(Throwable throwable, Response response) {
...
+    @Test
+    public void getTagsAsync() {
+        // Setup
+        CountDownLatch latch = new CountDownLatch(1);
+
+        // When
+        asyncClient.getBlobTags(containerName, blobName, new CallbackWithHeader<Map<String, String>, BlobGetTagsHeaders>() {
+                @Override
+                public void onSuccess(Map<String, String> result, BlobGetTagsHeaders header, Response response) {
+                    assertEquals(200, response.code());
```
vcolin7, line 1055: Should we check that the result's contents are correct like we check the body contents in `rawRownloadAsync()`?

vcolin7 (commented):
In general it LGTM, just had a couple questions :)

sdk/storage/azure-storage-blob/src/test/java/com/azure/android/storage/blob/BlobTest.java
rickle-msft: I think we have sync tests that validate the header values, and that logic should be shared between sync and async, so we didn't feel the need to test that redundantly. The async test is more to check that the request goes in and out correctly. Let me know what you think about that approach.

rickle-msft (approved):

sdk/storage/azure-storage-blob/src/test/java/com/azure/android/storage/blob/BlobTest.java
vcolin7: Sounds good to me as long as we have good coverage :)

vcolin7 (approved):

PR #450, opened 2020-11-09
Added options bags to storage container APIs

gapra-msft:


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/options/ContainerCreateOptions.java
```diff
@@ -0,0 +1,115 @@
...
+    private CancellationToken cancellationToken;
+
+    /**
+     * @param containerName The container name.
+     */
+    public ContainerCreateOptions(@NonNull String containerName) {
+        Objects.requireNonNull(containerName);
+        this.containerName  = containerName;
+    }
+
```
anuchandy, line 32: probably its wip, but just want to point Javadoc is in-complete for ctr, getters and setters

gapra-msft: Not sure if you commented this before or after I added some docs. Do you want me to add more detail than exists now? This docs are basically copy pasted from the Java SDK

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -91,38 +95,85 @@ public String getBlobServiceUrl() {
     /**
      * Creates a new container within a storage account. If a container with the same name already exists, the operation
      * fails.
+     * For more information, see the <a href="https://docs.microsoft.com/rest/api/storageservices/create-container">Azure Docs</a>.
      *
      * @param containerName The container name.
      */
-    public Void createContainer(String containerName) {
-        return storageBlobServiceClient.createContainer(containerName);
+    @NonNull
```
gapra-msft, line 102: is Void technically non-null? I'm not sure

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
```diff
@@ -91,38 +95,85 @@ public String getBlobServiceUrl() {
...
+     * Returns the container's metadata and system properties. For more information, see the
+     * <a href="https://docs.microsoft.com/rest/api/storageservices/get-container-metadata">Azure Docs</a>.
+     *
+     * @param options {@link ContainerGetPropertiesOptions}
+     * @return The response information returned from the server when getting a container's properties.
+     */
+    @NonNull
+    public ContainerGetPropertiesResponse getContainerPropertiesWithResponse(@NonNull ContainerGetPropertiesOptions options) {
+        Objects.requireNonNull(options);
+        BlobRequestConditions requestConditions = options.getRequestConditions() == null ? new BlobRequestConditions()
```
gapra-msft, line 172: Should we check that only lease id is set here?

gapra-msft: Related: I think the param validation should move from the impl class to this one

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
vcolin7: I'm pretty sure the only value a Void variable can have is null, so I wouldn't consider annotating it with `@NonNull`.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java
vcolin7: I don't think it matters that much if the required parameter validation happens here or in Impl, but the latter would mean we don't have to duplicate code I guess. Is there a particular reason you think that should be here? I think that if we are thinking the Impl class is going to be auto-generated then it makes sense to have validations outside of it.

vcolin7: Question: is `leaseId` the only required parameter in the options bag? Is `containerName` a part of it as well?

gapra-msft: lease Id is the only other service related optional param for this API.

gapra-msft: Yeah, I think since it will eventually be generated, we should do param validation on the client

vcolin7: If `leaseId` is required then I would add it to the options bag constructor and validate that its not null there instead of on the method in the client.

gapra-msft: leaseId is optional. Sorry if I mis-communicated before.

vcolin7: No problem. Even if `leaseId` is the only other value the service expects, since its optional I don't think we should validate its contents.

gapra-msft: We're not validating it's contents, just making sure the RequestConditions object exists before we call requestConditions.getLeaseId

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
```diff
@@ -259,43 +262,89 @@ public void cancel(Context context, long transferId) {
...
-    public void createContainer(String containerName,
-                                Integer timeout,
-                                Map<String, String> metadata,
-                                PublicAccessType publicAccessType,
-                                CancellationToken cancellationToken,
-                                CallbackWithHeader<Void, ContainerCreateHeaders> callback) {
+    public void createContainer(@NonNull ContainerCreateOptions options,
+                                @Nullable CallbackWithHeader<Void, ContainerCreateHeaders> callback) {
+        Objects.requireNonNull(options);
+        storageBlobServiceClient.createContainer(options.getContainerName(), options.getTimeout(),
```
rickle-msft, line 286: Why expand the options here rather than in the Intern method?

rickle-msft (approved):

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
gapra-msft: It's future proofing for when we generate code. The intern methods will eventually be generated so we probably want to do all this at the client level

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
```diff
@@ -44,14 +45,16 @@
...
 import com.azure.android.storage.blob.models.DeleteSnapshotsOptionType;
+import com.azure.android.storage.blob.models.ListBlobFlatSegmentHeaders;
 import com.azure.android.storage.blob.models.ListBlobsFlatSegmentResponse;
 import com.azure.android.storage.blob.models.ListBlobsIncludeItem;
 import com.azure.android.storage.blob.models.ListBlobsOptions;
 import com.azure.android.storage.blob.models.RehydratePriority;
-import com.azure.android.storage.blob.models.PublicAccessType;
+import com.azure.android.storage.blob.options.ContainerCreateOptions;
+import com.azure.android.storage.blob.options.ContainerDeleteOptions;
+import com.azure.android.storage.blob.options.ContainerGetPropertiesOptions;
```
vcolin7, line 57: Is it a common pattern to name the options bags with <Subject><Verb><Options>? I see we have other types like `ListBlobsOptions` so I'm not sure what's common in Java land.

vcolin7 (approved):
Other than my only comment, this looks good to me :)

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
gapra-msft: Yeah theres a design guideline to name the options bag <ClientType><Action>Options

PR #460, opened 2020-11-17
Added support for options bags in blobs

gapra-msft:
TODO: create options bags for getBlobsInPage
A little unclear on what the required and optional params are for that one though.

Should I just copy Java?

### DIFF
--- a/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
+++ b/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
@@ -314,6 +314,7 @@ public class StorageBlobAsyncClient {
     public void deleteContainer(@NonNull ContainerDeleteOptions options,
                                 @Nullable CallbackWithHeader<Void, ContainerDeleteHeaders> callback) {
         Objects.requireNonNull(options);
+        ModelHelper.validateRequestConditions(options.getRequestConditions(), false, true, true, false);
         storageBlobServiceClient.deleteContainer(options.getContainerName(), options.getTimeout(),
             options.getRequestConditions(), options.getCancellationToken(), callback);
     }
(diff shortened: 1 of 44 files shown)


sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobAsyncClient.java
```diff
@@ -423,205 +430,151 @@ public void onFailure(Throwable throwable, Response response) {
...
-                            String blobName,
-                            AccessTier tier,
-                            CallbackWithHeader<Void, BlobSetTierHeaders> callback) {
-        storageBlobServiceClient.setBlobTier(containerName,
-            blobName,
-            tier,
-            callback);
+    public void setBlobAccessTier(@NonNull String containerName,
+                                  @NonNull String blobName,
+                                  @Nullable AccessTier tier,
```
rickle-msft, line 547: I think tier is required and should be NonNull

gapra-msft: Right. Thanks for catching that

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/options/BlobGetPropertiesOptions.java
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Extended options that may be passed when getting properties of a blob.
+ */
+@Fluent
+public class BlobGetPropertiesOptions {
+
+    private final String containerName;
+    private final String blobName;
+    private String snapshot; /*  TODO: (gapra) : Should  we even expose this since we cant make snapshots anyway?*/
```
rickle-msft, line 24: I think yes because you can list and then read snapshots

bsiegel:
Adding related issue #316.

We've merged work to fix #314, so this PR will need to be rebased and have the newly-introduced types inherit from those introduced to azure-core in PR #511.

anuchandy:
closing this for now, please feel free to rebase if we still need these changes.

PR #476, opened 2020-12-03
Add Dokka docs configuration

bsiegel:
This generates HTML documentation using Dokka.

~It does not currently hook into the javadoc jar production process, that is still TBD.~
EDIT: This is now implemented. This PR should be ready to go.


sdk/core/azure-core/src/main/java/com/azure/android/core/http/package-info.java
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/**
+ * Common HTTP pipeline used by service clients.
```
vcolin7, line 5:
```suggestion
 * Common HTTP pipeline components used by service clients.
```

sdk/storage/azure-storage-blob/build.gradle
```diff
@@ -17,7 +17,6 @@ android {
 
     sourceSets {
         androidTest.assets.srcDirs += files("$projectDir/schema".toString())
-        test.java.srcDirs += files("${project(':sdk:core:azure-core').projectDir}/src/test/java/com/azure/android/core/common")
```
vcolin7, line 20: Did we figure out a way to share test classes from Azure Core?

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/credential/package-info.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+// Code generated by Microsoft (R) AutoRest Code Generator.
+
+/**
+ * Credential objects used with Azure Blob storage service clients to authorize requests.
```
vcolin7, line 6: Should this be capitalized like what we do with Azure Communication Services?

```suggestion
 * Credential objects used with Azure Blob Storage Service clients to authorize requests.
```

settings.gradle
```diff
@@ -1,6 +1,6 @@
...
-include ":sdk:core:azure-core"
-include ":sdk:storage:azure-storage-blob"
-include ":sdk:communication:azure-communication-common"
-include ":sdk:template:azure-sdk-template"
-include ":samples:sample-app-storage"
+include ':sdk:communication:azure-communication-common'
+include ':sdk:core:azure-core'
+include ':sdk:storage:azure-storage-blob'
+include ':sdk:template:azure-sdk-template'
+include ':samples:sample-app-storage'
```
vcolin7, line 6: Are we planning on using single quotes across our Gradle files?

sdk/storage/azure-storage-blob/build.gradle
bsiegel: Yep. I copied them over. This works, but if you re-import the project into Android Studio it no longer understands the project structure of Core because it mistakenly thinks that the src directory belongs to azure-storage-blob. Building from the commandline still works fine obviously, it's just a mistake in how Studio infers the project structure from the gradle files. I wasn't able to correct it either, "modules" section of the project structure dialog didn't show any of the libraries any more. After a couple hours knocking my head against the desk I figured out this line was causing the issue.

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/credential/package-info.java
bsiegel: Maybe? Even the official docs page mixes the casing: https://azure.microsoft.com/en-us/services/storage/blobs/

But I think it makes sense for consistency. I will correct.

settings.gradle
bsiegel: Generally I follow the advice of Ruby which also has this distinction (single quotes are literals, double quotes allow interpolation) and use single quotes everywhere I don't intend for interpolation to occur. I only changed them here because I was already re-ordering them. I considered changing them everywhere but felt that was too disruptive for this PR.

sdk/core/azure-core/src/main/java/com/azure/android/core/http/package-info.java
bsiegel: Resolved in d1f274a70

sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/credential/package-info.java
bsiegel: Resolved in d1f274a70

vcolin7 (approved):

PR #499, opened 2020-12-15
[Communication]: Add Identifier surfix to all identifier classes in communication-common

JianpingChen:
Add Identifier surfix to all identifier classes according to Azure board review

### DIFF
--- a/sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationUserIdentifier.java
+++ b/sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationUserIdentifier.java
@@ -8,17 +8,17 @@ import com.azure.android.core.util.CoreUtil;
 /**
  * Communication identifier for Communication Services Users
  */
-public class CommunicationUser extends CommunicationIdentifier {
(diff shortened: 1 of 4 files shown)


sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationUserIdentifier.java
```diff
@@ -8,17 +8,17 @@
 /**
  * Communication identifier for Communication Services Users
  */
-public class CommunicationUser extends CommunicationIdentifier {
+public class CommunicationUserIdentifier extends CommunicationIdentifier {
```
angiurgiu, line 11: Is this changing in Java too? There it would need changes to Chat as well, since the Builder.credential takes CommunicationUser

JianpingChen: Yes. The same change is happening in Java as well. c.f.
https://github.com/Azure/azure-sdk-for-java/pull/18064

marche0133 (approved):

PR #511, opened 2021-01-08
Added Core Options types: ClientOptions, RequestOptions and TransportOptions

### LINKED ISSUE #314: Azure Core Options types

vcolin7:
Fixes #314.

`TransportOptions` includes a `timeout` property, which won't be used yet by our `CancellationToken` pending some discussions about how to properly implement such a feature.

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.0.0-beta.3 (Unreleased)
 
+### New Features
+
+- Added `ClientOptions`, `RequestOptions` and `TransportOptions`.
+
 ## 1.0.0-beta.2 (2020-10-05)
 
 ### New Features
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/http/options/ClientOptions.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/http/options/ClientOptions.java
@@ -1,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.core.http.options;
+
+import androidx.annotation.NonNull;
+
+import com.azure.android.core.http.ServiceClient;
+import com.azure.android.core.util.logging.ClientLogger;
+
+/**
+ * Options for configuring a {@link ServiceClient}.
+ */
+public class ClientOptions {
+    @NonNull
+    private final String apiVersion;
+    @NonNull
+    private final ClientLogger logger;
+    @NonNull
+    private final TelemetryOptions telemetryOptions;
+    @NonNull
+    private final TransportOptions transportOptions;
+
+    /**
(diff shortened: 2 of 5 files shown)


vcolin7:
Not sure if we should rename `apiVersion` to `serviceVersion` instead, it appears that's what it's called in other languages. What do you think @bsiegel?

bsiegel:
I think I agree, `apiVersion` seems like the correct name.

vcolin7:
> 
> 
> I think I agree, `apiVersion` seems like the correct name.

Did you mean to say `serviceVersion`?

vcolin7:
Actually, it seems that the use of `serviceVersion` and `apiVersion` is split between languages, for exmaple, for KeyVault .NET and Java use the former and Python and JS use the latter.

vcolin7:
I think we can go with `apiVersion` for now to be consistent with iOS and change it later if we think it's necessary :) @bsiegel @anuchandy

bsiegel:
This looks good for Core, however it doesn't fully complete #314 - that requires that we also update client libraries to use this hierarchy. As it is now, I think this is fine but it means if we compile the entire SDK locally it will fail because clients will not be updated to the new options types. Do you want to try to address that here, or should we file an issue to update the client libraries as part of a followup?

bsiegel (approved):

vcolin7:
As discussed separately with @bsiegel, this change is additive for the most part and should not affect client libraries that depend on Azure Core.

PR #518, opened 2021-01-13
Add a default user agent for Android SDK

marche0133:


sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -206,6 +214,17 @@ public ChatAsyncClient build() {
...
             }
+            if (userAgentInterceptor == null) {
+                userAgentInterceptor = new UserAgentInterceptor(
+                    null,
+                    "azure-communication-chat",
+                    "1.0.0-beta.3",
+                    null,
+                    null,
+                    null);
+            }
```
JianpingChen, line 225: I believe there is standard way of formatting the agent string, c.f. https://azure.github.io/azure-sdk/general_azurecore.html and search for "User-Agent header"

JianpingChen (changes requested):

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
bsiegel: You should use the 3 or 4-parameter versions of the constructor for this, and make sure to pass a valid `Context`.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -170,6 +176,14 @@ public Builder credentialInterceptor(Interceptor credentialInterceptor) {
             return this;
         }
 
+        public Builder userAgentInterceptor(Context context) {
```
bsiegel, line 179: This should not be a public method on the builder. The builder should always create and insert all policies into the pipeline on behalf of the user - in most cases the user is not expected to understand what a "policy" is. Any policy that requires user-provided data should accept that data via a builder method, rather than accepting a policy instance. This also goes for things like `credentialInterceptor`, you should accept various 'credential' objects rather than an Interceptor.

My suggestion is to accept a Context object in the builder's constructor, and use it to create a UserAgentInterceptor which you then assign to a field. When the `build()` method is called, insert the UserAgentInterceptor in the appropriate place in the pipeline. See an example here: https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/StorageBlobClient.java#L791

bsiegel: For testing purposes you can also create a package-private constructor for the builder that accepts whatever you want (e.g. one or multiple interceptors, or Provider instances which is our abstraction around the Context - see here: https://github.com/Azure/azure-sdk-for-android/tree/master/sdk/core/azure-core/src/main/java/com/azure/android/core/provider and an example of them being used here: https://github.com/Azure/azure-sdk-for-android/blob/master/sdk/core/azure-core/src/test/java/com/azure/android/core/http/interceptor/UserAgentInterceptorTest.java#L289

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatConfig.java
```diff
@@ -0,0 +1,6 @@
+package com.azure.android.communication.chat;
+
+public class ChatConfig {
```
bsiegel, line 3: Recommend making this package-private unless you really intend to expose these constants to the user.

bsiegel (changes requested):

PR #557, opened 2021-02-05
[Communication]: Identifier serializer update for new model

JianpingChen:


sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationIdentifierSerializer.java
```diff
@@ -3,77 +3,106 @@
...
+    /**
+     * Deserialize CommunicationIdentifierModel into CommunicationIdentifier
+     * @param identifier CommunicationIdentifierModel to be deserialized
+     * @return deserialized CommunicationIdentifier
+     */
     public static CommunicationIdentifier deserialize(CommunicationIdentifierModel identifier) {
-        String id = identifier.getId();
-        CommunicationIdentifierKind kind = identifier.getKind();
+        assertSingleType(identifier);
+        String rawId = identifier.getRawId();
```
RezaJooyandeh, line 19: Do not we need a `requireNonNull` here as well?

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationUserIdentifier.java
```diff
@@ -15,7 +15,7 @@
     /**
      * Creates a CommunicationUserIdentifier object
      *
-     * @param id the string identifier representing the identity
+     * @param id id of the communication user.
```
RezaJooyandeh, line 18: >id id of

➡ "id of"

RezaJooyandeh (approved):

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationUserIdentifier.java
JianpingChen: The first "id" is the parameter name. What about "@param id identifier string of the Communication user"

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationIdentifierSerializer.java
JianpingChen: To my understanding, rawId is only required for TeamsUser and UnknownUser. For phone number and communication user, it is optional. Am I missing anything?

RezaJooyandeh: In the current schema `rawId` is required only for those two, but as a contract we want the services to always return it. So let's make it mandatory in the deserialization

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationUserIdentifier.java
RezaJooyandeh: oh I had misunderstood how the comments work in Java. I think we are good.

PR #567, opened 2021-02-10
Create signaling package to integrate Android chat SDK with trouter client to receive realtime messages

marche0133:
Create signaling package to integrate Android chat SDK with trouter client.
Receive real-time notifications through trouter.
The payloads and event classes are not used, they are only for reference purpose.

Others:
To use signaling package, we need to import trouter client package as sdk dependency, trouter team is currently working on it to release trouter so I can update gradle to include this dependency. For now, the build will fail because of lacking the dependency, but it is fine to merge it to this feature branch.
Tech brief is here, https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/15497/Android-Trouter-Integration-Tech-Brief

TODO:
A demo app will be added in future


sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
```diff
@@ -0,0 +1,343 @@
...
+            }
+        } catch (JSONException e) {
+            System.out.println(e.getMessage());
+        }
+
+        return toEventPayload(chatEventId, genericPayload);
+    }
+
+    public static JSONObject toEventPayload(String eventId, JSONObject payload) {
+        if (eventId.equals(ChatEventId.chatMessageReceived.name())) {
```
vcolin7, line 50: Since the value of `eventId` does not change and this method will only enter one of the many if blocks below, wouldn't it make more sense to modularize this logic into smaller methods like `toChatMessageReceivedPayload` or `createChatMessageReceivedPayload` use a switch to call the right one based on the `eventId` contents?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/CommunicationListener.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public CommunicationListener(String chatEventId, RealTimeNotificationCallback listener) {
+        this.chatEventId = chatEventId;
+        this.listenerFromConsumer = listener;
+    }
+
+    @Override
+    public void onTrouterConnected(String endpointUrl, ITrouterConnectionInfo connectionInfo) {
+        final String msg = "onTrouterConnected(): url=" + endpointUrl + ", newPublicUrl=" + Boolean.toString(connectionInfo.isNewEndpointUrl());
+        System.out.println(msg);
```
vcolin7, line 24: I don't think we want to be printing directly to the system console. We should consider updating this to use the Azure Core logger or some other logging mechanism.

vcolin7: Then same goes for all other places where we use `System.out.something`.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/CommunicationSignalingClient.java
```diff
@@ -0,0 +1,148 @@
...
+            platformUiVersion,
+            testTemplateKey,
+            null,
+            ""
+        );
+        final TrouterUrlRegistrar registrar = new TrouterUrlRegistrar(
+            skypetokenProvider,
+            registrationData,
+            defaultRegistrarHostnameAndBasePath,
+            3600
```
vcolin7, line 80: We should move this to a constant.

vcolin7 (commented):
Added a few suggestions. I assume we will add Javadoc comments on everything before merging to master, correct? Also, should we consider moving the `chatevents` and `properties` packages under a `models` package? What do you think @anuchandy? Finally, we should add the licensing header to all our files :)

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -155,6 +164,62 @@ public void listChatThreadsNext(String nextLink, final Callback<Page<ChatThreadI
...
+        }
+
+        this.isRealtimeNotificationsStarted = false;
+        this.signalingClient.stop();
+    }
+
+    /**
+     * Listen to a chat event.
+     */
+    public void on(String chatEventId, RealTimeNotificationCallback listener) {
```
anuchandy, line 198: Is this api for attaching a listener for real time notification? I wonder we can come-up with a more specialized name instead of the generic name `on`. (the same comment for `off` method as well).

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -214,6 +285,16 @@ public Builder userAgentInterceptor(Interceptor userAgentInterceptor) {
...
 
+        public Builder context(Context context) {
+            this.context = context;
+            return this;
+        }
+
+        public Builder userToken(String userToken) {
+            this.userToken = userToken;
+            return this;
+        }
```
anuchandy, line 296: I see, so if user set `context` and `userToken` then we create `SignalingClient` which enables real time notification, I feel like, it might be bit hard for users to discover this relationship, what do you think?

How about a builder method name that better communicate this? Something like `enableRealtimeNotification(context, userToken)`

anuchandy (commented):
will continue to review this, just sharing initial comments,

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
anuchandy: something like `[add|set]RealtimeNotificationCallback`?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/SignalingClient.java
```diff
@@ -0,0 +1,20 @@
+package com.azure.android.communication.chat.signaling;
+
+public interface SignalingClient {
```
anuchandy, line 3: any reason to have the public contract `SignalingClient`? are we expecting chat module to ship with more clients (in addition to `CommunicationSignalingClient`) that implements this?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/CommunicationListener.java
```diff
@@ -0,0 +1,66 @@
+package com.azure.android.communication.chat.signaling;
+
+import com.microsoft.trouterclient.ITrouterConnectionInfo;
+import com.microsoft.trouterclient.ITrouterListener;
+import com.microsoft.trouterclient.ITrouterRequest;
+import com.microsoft.trouterclient.ITrouterResponse;
+
+import org.json.JSONObject;
+
+public class CommunicationListener implements ITrouterListener {
```
anuchandy, line 10: just double checking, do we want to expose this as a public type? If the end user is not suppose to use this directly then we can either make it package private or move it to implementation package.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/CommunicationSignalingClient.java
```diff
@@ -0,0 +1,148 @@
...
+import static com.azure.android.communication.chat.signaling.SignalingConfig.platformUiVersion;
+import static com.azure.android.communication.chat.signaling.SignalingConfig.testApplicationId;
+import static com.azure.android.communication.chat.signaling.SignalingConfig.testTemplateKey;
+import static com.azure.android.communication.chat.signaling.SignalingConfig.trouterClientVersion;
+
+public class CommunicationSignalingClient implements SignalingClient {
+    private TrouterClientHost trouterClientHost;
+    private ISelfHostedTrouterClient trouter;
+    String userToken;
+    Set<CommunicationListener> trouterListeners;
```
anuchandy, line 30: are we accessing `trouterListeners` and `userToken` outside of this class but within the package? just making sure its package private visibility is intentional

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/RealTimeNotificationCallback.java
```diff
@@ -0,0 +1,9 @@
+package com.azure.android.communication.chat.signaling;
+
+import org.json.JSONObject;
+
+public interface RealTimeNotificationCallback {
+
+    public void onChatEvent(JSONObject chatEvent);
```
anuchandy, line 7: If we are not expecting to add more methods to this interface, then we can mark it as `FunctionalInterface` so that end user can use lambda.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/SignalingConfig.java
```diff
@@ -0,0 +1,15 @@
+package com.azure.android.communication.chat.signaling;
+
+public class SignalingConfig {
```
anuchandy, line 3: Is this a type we want to expose to the end user? if not we can make it package private or move to implementation package.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
```diff
@@ -0,0 +1,343 @@
...
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
+public class TrouterUtils {
```
anuchandy, line 14: If this is a type only used internally and should not be exposed to the end user, then consider making it package private or move to implementation package.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterNotificationPayload.java
```diff
@@ -0,0 +1,101 @@
...
+    String originalArrivalTime;
+}
+
+class MessageEditedPayload extends BasePayload {
+    String messageId;
+    String clientMessageId;
+    String senderDisplayName;
+    String messageBody;
+    String version;
+    String edittime;
```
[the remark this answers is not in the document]
marche0133: Can't change it since it is what trouter client outputs to us.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterNotificationPayload.java
```diff
@@ -0,0 +1,101 @@
...
+    String originalArrivalTime;
+}
+
+class MessageDeletedPayload extends BasePayload {
+    String messageId;
+    String clientMessageId;
+    String senderDisplayName;
+    String version;
+    String deletetime;
```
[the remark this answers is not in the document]
marche0133: Same here

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterNotificationPayload.java
```diff
@@ -0,0 +1,101 @@
...
+
+class ReadReceiptReceivedPayload extends BasePayload {
+    String messageId;
+    String clientMessageId;
+    String messageBody;
+}
+
+class ReadReceiptMessageBody {
+    String user;
+    String consumptionhorizon;
```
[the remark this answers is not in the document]
marche0133: Same here

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
```diff
@@ -0,0 +1,343 @@
...
+        put(ChatEventId.chatMessageReceived, 200);
+        put(ChatEventId.typingIndicatorReceived, 245);
+        put(ChatEventId.readReceiptReceived, 246);
+        put(ChatEventId.chatMessageEdited, 247);
+        put(ChatEventId.chatMessageDeleted, 248);
+        put(ChatEventId.chatThreadCreated, 257);
+        put(ChatEventId.chatThreadPropertiesUpdated, 258);
+        put(ChatEventId.chatThreadDeleted, 259);
+        put(ChatEventId.participantsAdded, 260);
+        put(ChatEventId.participantsRemoved, 261);
```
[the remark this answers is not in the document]
marche0133: When we receive events from trouter client, there will be a code in the payload. I get this listing from our js sdk signaling package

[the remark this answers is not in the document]
vcolin7: We should probably have an enum or constants for these as opposed to hardcoding them. I feel that someone reading this code might get confused as to where they come from.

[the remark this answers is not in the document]
vcolin7: Not a blocker for now, in my opinion, but something to keep in mind.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterNotificationPayload.java
```diff
@@ -0,0 +1,101 @@
+package com.azure.android.communication.chat.signaling;
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+class BasePayload {
```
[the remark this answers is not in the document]
marche0133: Users don't need to know much about Trouter related stuff, so I just put those payloads in single file but for chat events I would like to put them in separate files.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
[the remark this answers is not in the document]
marche0133: I will add a java doc to explain they are trouter id codes.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
marche0133: I just follow other languages, they put in one method instead of creating methods.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -41,12 +46,16 @@
...
+    private boolean isRealtimeNotificationsStarted;
 
     /**
      * Initializes an instance of Chats client.
      */
-    ChatAsyncClient(ChatsImpl serviceClient) {
+    ChatAsyncClient(ChatsImpl serviceClient, SignalingClient signalingClient) {
         this.serviceClient = serviceClient;
+        this.signalingClient = signalingClient;
+        this.isRealtimeNotificationsStarted = false;
```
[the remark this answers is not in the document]
marche0133: Yes, currently we already have some changes that need to do manually. I will see if we can make this part automatic in future, not sure though.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
marche0133: We explicitly call it on and off in other languages, so I will have to keep it.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/SignalingClient.java
marche0133: It needs to be public so that chatClient can call methods of it.

vcolin7:
Remember we have to add the License header to all the files in the repository :)

marche0133:
> Remember we have to add the License header to all the files in the repository :)

Yes I have added the header to files, did you see some file still missing header?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
vcolin7: I understand that but I think we should strive to make code as best as we can. I think it's better for readability and maintainability to avoid methods this big that do only one of many things based on a single condition. The fact that other languages do it does not necessarily mean that it's the best way to do things. It's not a blocker to approve this PR, but I think it's important to keep this in mind.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/properties/ChatEventId.java
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.communication.chat.signaling.properties;
+
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+
```
vcolin7, line 9: Removing additional header.

```suggestion
package com.azure.android.communication.chat.signaling.properties;

```

vcolin7 (commented):
I am not entirely familiar with the process of using an internal library like what you are doing here. Is it internal because it's closed-source? Is it safe to include the AAR in the repository?

vcolin7:
> 
> 
> > Remember we have to add the License header to all the files in the repository :)
> 
> Yes I have added the header to files, did you see some file still missing header?

Last time I checked I saw something like that, it is possible my browser had cached the webpage showing the file in a previous state, so nothing to worry about.

PR #635, opened 2021-03-09
[Communication] - Common - CommunicationCloudEnvironment Fix

minnieliu:


sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationCloudEnvironment.java
```diff
@@ -7,36 +7,45 @@
...
+    public CommunicationCloudEnvironment() {
+        this.environmentValue = PUBLIC_VALUE;
+    }
+
+    /**
+     * Sets an environment value from a String
+     * @param environmentValue string representation of the environment value
+     * @return CommunicationCloudEnvironment object itself
+     */
+    public final CommunicationCloudEnvironment fromString(String environmentValue) {
```
RezaJooyandeh, line 29: We have a feedback on the Java ApiView to make this `static`.

JianpingChen (approved):
Please update the APIReview if you can. Otherwise, please let me know an I'll update it.

https://apiview.dev/Assemblies/Review/1dad7ec250684647aaeb2b17c59ad7e3

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationCloudEnvironment.java
minnieliu: @RezaJooyandeh  If we make this static, then we cannot refer to "this" in the function implementation. Then should we be changing environmentValue to static as well?

JoshuaLai (approved):
lgtm

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationCloudEnvironment.java
RezaJooyandeh: No, essentially this would be a static method that instantiate a new `CommunicationCloudEnvironment` based in the input and returns it.

PR #681, opened 2021-03-22
Integrating Chat preview-5 features, adding ChatThread builder and addressing arch review

CHILIU-MSFT:
Updates:

1. Rename ChatThread model to ChatThreadProperties
2. Rename GetChatThread* operation to GetProperties*
3. Add ChatThreadClientBuilder
4. Update ChatThreadClient.AddParticipant -  throw ChatError instead of return the invalid user list
5. Rename parameter in CreateChatThreadOption  repeatabilityRequestId =>idempotencyToken
6. Update in ChatThreadClient SendMessage : Use SendChatMessageResult instead of string (messageId)


sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadAsyncClient.java
```diff
@@ -223,11 +225,35 @@ public String getChatThreadId() {
...
+     */
+    CompletableFuture<Response<AddChatParticipantsResult>> addParticipant(AddChatParticipantsOptions options,
+                                                                           Context context) {
+        context = context == null ? Context.NONE : context;
+        return this.addParticipants(options, context)
+            .thenApply(result -> {
+                if (result.getValue().getInvalidParticipants() != null) {
+                    if (result.getValue().getInvalidParticipants().size() > 0) {
+                        ChatError error = result.getValue().getInvalidParticipants().get(0);
+                        this.logger.logExceptionAsError(new InvalidParticipantException(error));
```
CHILIU-MSFT, line 250: To be honest, I don't have much experience on Android dev. I am not sure whether this is the proper way for throwing a exception. Do you have any suggestions?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/InvalidParticipantException.java
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT License.
+
+package com.azure.android.communication.chat.models;
+
+import com.azure.android.core.exception.AzureException;
+
+/**
+ * Exception when a participant cannot be added to a chat thread
+ */
+public final class InvalidParticipantException extends AzureException {
```
JianpingChen, line 11: This will make SDK users error handling more complicated since now we have two base classes for excpetions: HttpResponseException and AzureException. We should use only one not both.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatThreadProperties.java
```diff
@@ -9,9 +9,9 @@
 
 import org.threeten.bp.OffsetDateTime;
 
-/** The ChatThread model. */
+/** The ChatThreadProperties model. */
 @Fluent
-public final class ChatThread {
+public final class ChatThreadProperties {
```
JianpingChen, line 14: This class looks the same to the one under implemenation/models. Can we just move that one to models folder instead? It is just a simple change to the README.md in swagger folder.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadAsyncClient.java
anuchandy: The return type composes the error field and has no other fields. If the expectation is to throw this composed error from here, then should we make the return type Void.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/CreateChatThreadOptions.java
```diff
@@ -21,10 +21,10 @@
     /*
      * Members to be added to the chat thread.
      */
-    @JsonProperty(value = "participants", required = true)
+    @JsonProperty(value = "participants")
     private List<ChatParticipant> participants;
 
-    private String repeatabilityRequestId;
+    private String idempotencyToken;
```
JianpingChen, line 27: How is this property useful if it is ignored by the CreateChatThreadOptionsConverter?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadAsyncClient.java
anuchandy: I see this is a common internal function used from two other places
1. `public CompletableFuture<Void> addParticipant(..)`
2. `public CompletableFuture<Response<AddChatParticipantsResult>> addParticipantWithResponse(..)`

Are you expecting not to throw from 2 but only from 1, then you you might want to do the check and throw only from 1.

If you expect to throw an exception from both 1 and 2, you might want to change the return type of 2 to `CF<Response<Void>>` and add check and throw.

CHILIU-MSFT: Correct. The expectation is to throw for both 1 and 2. I'll fix it.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatThreadProperties.java
anuchandy: I think the one in impl based on swagger has the field `CommunicationIdentifierModel` which we don't want to expose to the users. This public type exposes that identifier as `CommunicationIdentifier` (from common).

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/CreateChatThreadOptions.java
CHILIU-MSFT: idempotencyToken is in path (not) body of the operation. See line 104 of ChatAsyncClient.java.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/InvalidParticipantException.java
CHILIU-MSFT: Based on offline discussion, we would keep InvalidParticipantException extends AzureException

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -254,7 +262,27 @@ public ChatThreadAsyncClient getChatThreadClient(String chatThreadId) {
      */
     CompletableFuture<Response<Void>> deleteChatThread(String chatThreadId, Context context) {
         context = context == null ? Context.NONE : context;
-        return this.chatClient.deleteChatThreadWithResponseAsync(chatThreadId, context);
+        return this.chatClient.deleteChatThreadWithResponseAsync(chatThreadId, context)
+            .exceptionally(throwable -> translateException(throwable));
+    }
+
+    private <T> T translateException(Throwable throwable) {
```
JianpingChen, line 269: T is never used.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadAsyncClient.java
```diff
@@ -1008,4 +1046,23 @@ public String getChatThreadId() {
                     null);
             });
     }
+
+    private <T> T translateException(Throwable throwable) {
```
JianpingChen, line 1050: Duplicate code. Consider moving this to a converter and make it package private

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
CHILIU-MSFT: I use a template here because return type could be Response<CreateChatThreadResult>, PagedResponse<ChatThreadItem> or Response<Void>. Although in this function it's thrown only and there won't be anything returned, I am trying to find a solution to have one function to fit all those types. Do you have any idea for this?

JianpingChen: It'll be better if we can create a unit test for this chunk of code to fully understand the behaviour. From the documentation, it is not very clear to me whether we should throw or not.

In createChatThread case, there is another thenApply after exceptionally. If you return null here, will the logic inside thenApply being triggered with Completable<Null>? If yes, there will be a null pointer exception. To avoid that, we need to throw, IMHO.

JianpingChen: If we should throw, this function becomes 'Throwable translate(Throwable t)'. At the place it is used, 'exceptionally(throwable -> { throw translateException(throwable); });

CHILIU-MSFT: Good idea! I'll update.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatErrorResponseException.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.azure.android.communication.chat.models;
+
+import com.azure.android.core.http.exception.HttpResponseException;
+import com.azure.android.core.http.HttpResponse;
+
+/**
+ * Exception thrown on chat specific errors
+ */
+public class ChatErrorResponseException extends HttpResponseException {
```
anuchandy, line 12: we can mark it as a `final` class

anuchandy (approved):
thanks lgtm! (same feedback as JP wrt error handling)

JianpingChen (approved):

PR #691, opened 2021-03-24
Updating azure-communication-common module to be in sync with beta.7 (released from master) and additional improvements

anuchandy:
1. Syncing android azure-communication-common module with the same in master.
1.1 Removing of azure-core dependency.
1.2 Removing identifier-model types, serializers and it's tests.
2. Adding TokenParser tests.
3. Adding communication identifier tests
4. Return failed CF when try to acquire the token from a disposed credentials (consistent with java behavior)
5. Picking applicable apiView feedback from java common


sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClientBuilder.java
```diff
@@ -178,8 +178,8 @@ public ChatAsyncClient buildAsyncClient() {
             final HttpPipelinePolicy authorizationPolicy;
             if (this.communicationTokenCredential != null) {
                 authorizationPolicy = chain -> {
-                    final Future<AccessToken> tokenFuture = this.communicationTokenCredential.getToken();
-                    final AccessToken token;
+                    final Future<CommunicationAccessToken> tokenFuture = this.communicationTokenCredential.getToken();
```
minnieliu, line 181: Are we still using Future here instead of CompletableFuture?

sdk/communication/azure-communication-common/src/test/java/com/azure/android/communication/common/CommunicationTokenCredentialTest.java
```diff
@@ -212,17 +207,18 @@ public void getToken_onDemandAutoRefresh_noInitialToken_firstFetch_multithreaded
         // Set up blocked multithreaded calls
         Runnable blockedRefresh = this.arrangeBlockedRefresh(mockTokenRefresher);
         int numCalls = 5;
-        List<Future<AccessToken>> accessTokenFutures = new ArrayList<>();
+        List<Future<CommunicationAccessToken>> accessTokenFutures = new ArrayList<>();
```
minnieliu, line 210: CompletableFuture here instead of Future. Or is there a particular reason why we want to keep as Future?

minnieliu: I see instances of Future all throughout the file as well

minnieliu:
I created an API View based off of this branch and noticed that the version number is beta6 instead of beta8: https://apiview.dev/Assemblies/Review/1dad7ec250684647aaeb2b17c59ad7e3

Can you please update it? @anuchandy

sdk/communication/azure-communication-common/src/test/java/com/azure/android/communication/common/CommunicationTokenCredentialTest.java
anuchandy: yes, `CompletableFuture` is a `Future`, so unless we're exercising specific apis from CF in the tests, its totally fine to use receiving type as `Future`. But thinking a bit more, I feel that it's a good idea to use concrete type, just to catch any breaking in return type, will update the PR to `CF`.

JianpingChen:
Please also update Common ReadMe.md to reflect the change Future -> CompletableFuture

JianpingChen:
Also ChangeLog.md to reflect the breaking changes please

sdk/communication/azure-communication-common/build.gradle
```diff
@@ -12,9 +12,7 @@ android {
```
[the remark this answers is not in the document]
minnieliu: Yea, agree with JP, the version should probably be beta.9? 

Also noticed that in the README.md for common, it is still referring to beta.7. We should update that to be consistent as well

sdk/communication/azure-communication-common/README.md
```diff
@@ -38,13 +38,13 @@ Add an `implementation` configuration to the `dependencies` block of your app's
 // build.gradle
 dependencies {
     ...
-    implementation "com.azure.android:azure-communication-common:1.0.0-beta.5"
+    implementation "com.azure.android:azure-communication-common:1.0.0-beta.7"
```
minnieliu, line 41: Should this be beta.9 as JP suggested?

anuchandy: Readme should contain reference to the latest "released" version. For example if we look at "Java", you can see that readme is [referring](https://github.com/Azure/azure-sdk-for-java/tree/master/sdk/communication/azure-communication-common#include-the-package) the last released version in [maven](https://search.maven.org/artifact/com.azure/azure-communication-common).

Similarly for "Android" the latest "released" version is 1.0.0-beta.7 in [maven](https://search.maven.org/artifact/com.azure.android/azure-communication-common), hence Readme use this version.

Screen Shot 2021-03-24 at 11 27 23 AM

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClientBuilder.java
anuchandy: this is addressed.

JianpingChen (approved):

minnieliu (approved):

sdk/communication/azure-communication-common/build.gradle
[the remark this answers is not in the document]
anuchandy: I've updated the build.gradle to use beta-9 and ensured the changelog is correct. Regarding the readme update, I'll do the update once beta-8 is out from master, created a work item for [that](
 https://github.com/Azure/azure-sdk-for-android/issues/693)

PR #711, opened 2021-03-30
Update naming, add unit test for identifier parsing

CHILIU-MSFT:
Update naming
Add unit test for identifier parsing

### DIFF
--- a/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
+++ b/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
@@ -114,29 +114,29 @@ class TrouterUtils {
         }
     }
 
-    public static CommunicationIdentifier getCommunicationIdentifier(String mri) {
-        if (mri.startsWith(TEAMS_PUBLIC_USER_PREFIX)) {
-            return new MicrosoftTeamsUserIdentifier(mri.substring(TEAMS_PUBLIC_USER_PREFIX.length()), false)
-                .setRawId(mri)
+    public static CommunicationIdentifier getCommunicationIdentifier(String rawId) {
+        if (rawId.startsWith(TEAMS_PUBLIC_USER_PREFIX)) {
+            return new MicrosoftTeamsUserIdentifier(rawId.substring(TEAMS_PUBLIC_USER_PREFIX.length()), false)
+                .setRawId(rawId)
                 .setCloudEnvironment(CommunicationCloudEnvironment.PUBLIC);
-        } else if (mri.startsWith(TEAMS_DOD_USER_PREFIX)) {
-            return new MicrosoftTeamsUserIdentifier(mri.substring(TEAMS_DOD_USER_PREFIX.length()), false)
-                .setRawId(mri)
+        } else if (rawId.startsWith(TEAMS_DOD_USER_PREFIX)) {
+            return new MicrosoftTeamsUserIdentifier(rawId.substring(TEAMS_DOD_USER_PREFIX.length()), false)
+                .setRawId(rawId)
                 .setCloudEnvironment(CommunicationCloudEnvironment.DOD);
-        } else if (mri.startsWith(TEAMS_GCCH_USER_PREFIX)) {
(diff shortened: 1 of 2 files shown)


sdk/communication/azure-communication-chat/src/test/java/com/azure/android/communication/chat/signaling/TrouterUtilsTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+        assertNotNull(communicationIdentifier);
+        assertTrue(communicationIdentifier instanceof CommunicationUserIdentifier);
+        CommunicationUserIdentifier communicationUserIdentifier =
+            (CommunicationUserIdentifier) communicationIdentifier;
+        assertEquals(acsUserMri, communicationUserIdentifier.getId());
+    }
+
+    @Test
+    public void canParseSpoolUserMri() {
```
RezaJooyandeh, line 96: Why do we still have 8:spool here? These must be retired, are not they?

sdk/communication/azure-communication-chat/src/test/java/com/azure/android/communication/chat/signaling/TrouterUtilsTest.java
```diff
@@ -0,0 +1,132 @@
...
+        MicrosoftTeamsUserIdentifier microsoftTeamsUserIdentifier =
+            (MicrosoftTeamsUserIdentifier) communicationIdentifier;
+        assertEquals(CommunicationCloudEnvironment.PUBLIC, microsoftTeamsUserIdentifier.getCloudEnvironment());
+        assertTrue(microsoftTeamsUserIdentifier.isAnonymous());
+        assertEquals(USER_ID, microsoftTeamsUserIdentifier.getUserId());
+        assertEquals(teamsVisitorUserMri, microsoftTeamsUserIdentifier.getRawId());
+    }
+
+    @Test
+    public void canParseAcsUserMri() {
```
RezaJooyandeh, line 84: instead of MRI, let's call it RawId. This should make sense for a developer that is not familiar with our internal terminolgy

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
```diff
@@ -114,6 +114,32 @@ public static BaseEvent toEventPayload(ChatEventId eventId, JSONObject payload)
         }
     }
 
+    public static CommunicationIdentifier getCommunicationIdentifier(String mri) {
```
RezaJooyandeh, line 117: mri -> rawId

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
```diff
@@ -389,38 +420,12 @@ private static String parseShareHistoryTime(Long shareHistoryTimeEpochMilli) {
             .toString();
     }
 
-    private static String parseRecipientSkypeId(String skypeId) {
-        if (skypeId.startsWith(USER_PREFIX)) {
-            return skypeId;
+    private static String parseRecipientRawId(String rawId) {
```
RezaJooyandeh, line 423: I think I misunderstood what this function does originally. It seems to be the temporary workaround for when the prefix is missing in the id. If that is the case, you can call it `addRawIdPrefixIfMissing` which would explains what it does literarily.

sdk/communication/azure-communication-chat/src/test/java/com/azure/android/communication/chat/signaling/TrouterUtilsTest.java
CHILIU-MSFT: It should. But every time when I ask Jiri Sofka to check whether we've already deleted all spool resources created during private preview, he said this work is still not done yet. So, I can't say they are all retired...

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/signaling/TrouterUtils.java
RezaJooyandeh: And once we update the template, we can remove this

JianpingChen (approved):

PR #727, opened 2021-04-09
refactor signaling package

CHILIU-MSFT:
1) make signaling package private, only expose event models
2) re-use ChatParticipant & ChatThreadProperties
3) Use full MRI from notification payload instead of add "8:" in client


sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClient.java
```diff
@@ -7,8 +7,8 @@
 import com.azure.android.communication.chat.models.CreateChatThreadOptions;
 import com.azure.android.communication.chat.models.CreateChatThreadResult;
 import com.azure.android.communication.chat.models.ListChatThreadsOptions;
-import com.azure.android.communication.chat.signaling.RealTimeNotificationCallback;
-import com.azure.android.communication.chat.signaling.properties.ChatEventId;
+import com.azure.android.communication.chat.models.signaling.RealTimeNotificationCallback;
+import com.azure.android.communication.chat.models.signaling.ChatEventId;
```
anuchandy, line 11: Should we flatten these models instead of nesting? i.e. completely removing `signaling` from public surface

CHILIU-MSFT: That's an option as well. Would adding a signaling package make the code more readable?

RezaJooyandeh:
Do you mind updating the ApiView and share the link here too?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClient.java
anuchandy: We decided to merge notification APIs with ChatClient, so I was thinking, why we want to give user an impression that notification is something special even after the merge; For instance, user write the below code:

```java
communication.chat.ChatAsyncClient chatAsyncClient = ..

communication.chat.model.CreateChatThreadResult result = chatAsyncClient.createChatThread()..;
```

```java
chatAsyncClient.on(threadId, new communication.chat.model.signaling.RealTimeNotificationCallback() {
   ...
   communication.chat.model.signaling.BaseEvent = ..
});
```

from user side, he is  using a single client but some of the methods in the same client uses models from one namespace, some from nested model namespace, just wondering what value this separation brings.

CHILIU-MSFT:
https://apiview.dev/Assemblies/Review/479be75ee46a450484cbd286a0bb0983

@RezaJooyandeh  I uploaded ApiView (without merge package models.signaling & models)

anuchandy:
@CHILIU-MSFT when you get a chance could please sync with upstream master? (to get a clean CI build with link verification working)

RezaJooyandeh:
@CHILIU-MSFT I noticed in the ApiView that there a lot method about internal mechanic of pagination that are public. For example instead of `GetChatThreads`, we have `GetChatThreadFirstPage` and `GetChatThreadNextPage`.

CHILIU-MSFT:
@RezaJooyandeh GetChatThreads is replaced by GetChatThreadFirstPage and GetChatThreadNextPage in preview-5 rest API update before I take over Android.
@anuchandy @juancamilor Is this part of review feedback or Android principles? Do you have any idea? Thanks!

juancamilor:
> @RezaJooyandeh GetChatThreads is replaced by GetChatThreadFirstPage and GetChatThreadNextPage in preview-5 rest API update before I take over Android.
> @anuchandy @juancamilor Is this part of review feedback or Android principles? Do you have any idea? Thanks!

You can use JAVA as a reference. The comments / suggestions there should apply here as well 
https://apiview.dev/Assemblies/Review/76234f04f68a487ea8a0e168224435a7/cbaf15d6fc1c4457970aac68c79579ad?doc=False

It can be done incrementally as well ..

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatMessageDeletedEvent.java
```diff
@@ -12,28 +14,33 @@
      * The timestamp when the message was deleted. The timestamp is in ISO8601 format:
      * `yyyy-MM-ddTHH:mm:ssZ`.
      */
+    @JsonProperty(value = "deletedOn")
```
anuchandy, line 17: have we considered using `OffsetDataTime` for `deletedOn` property in `ChatMessageDeletedEvent` type?

anuchandy: [same comment for eligible date properties in other event models]

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatMessageDeletedEvent.java
```diff
@@ -12,28 +14,33 @@
...
     private String deletedOn;
 
     /**
      * The Id of the message. This Id is server generated.
      */
+    @JsonProperty(value = "id")
```
anuchandy, line 23: In case event types are output only (from service -> user) do we need `set*` methods anymore?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatMessageReceivedEvent.java
```diff
@@ -12,38 +14,45 @@
      * Type of the chat message.
      * The only type currently supported is Text
      */
+    @JsonProperty(value = "type")
```
anuchandy, line 17: do we want to consider an ext-enum for this?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatEventId.java
```diff
@@ -16,6 +16,5 @@
     chatThreadDeleted,
     chatThreadPropertiesUpdated,
     participantsAdded,
-    participantsRemoved,
-    connectionChanged;
+    participantsRemoved
```
anuchandy, line 19: the naming convention for enum value is - all caps with appropriate use of underscore.. e.g. `CHAT_MESSAGE_RECEIVED`

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatMessageReceivedEvent.java
RezaJooyandeh: +1 The public facing version of this (if there is a public version) should be extended enum. But if it is purely internal, then I think it is ok, right @anuchandy?

anuchandy: yes, it's fine for internal types. It seems this model is a part of public API, the model represents event emitted to the public `on` method

RezaJooyandeh: Yeah it seems that they are exposed through `ChatEventId` which should become an extended enum.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatEventId.java
RezaJooyandeh: Is not it better to call this `ChatEventKind`? It is not really the id of an event. Id implies identification and uniqueness.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatMessageDeletedEvent.java
CHILIU-MSFT: We need the setters when parsing this object from json from trouter.
The field names are different between json from trouter and what we expose to customer. There's a mapping in implementation/signaling/TrouterUtils.java

anuchandy: When we talked about using `JacksonSerder`, I thought it is to replace the parsing logic in `TrouterUtil` from `JsonObject` to `Jackson`, but it seems we use it only to simplify the logging experience in apps.

But now we can remove those "public setters (not usable for customers)," Jackson helps here; it does not require setter methods to set value for properties with "JsonProperty" annotation.

Special care is needed to populate those properties with the type "CommunicationIdentifier" in `ChatUserEvent`; since the wire representation is different, Jackson cannot set those properties. We can use an accessor pattern to solve that.

Here is a sample impl showing removal of unused setters from our public API and the accessor pattern in impl level; I did it only for `Deleted` and `Edited` event types but the concept is same across event models.

https://github.com/anuchandy/azure-sdk-for-android-1/commit/d7a280293714438a9c02a5a6166884b89382dfb5

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatEventKind.java
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.communication.chat.models;
+
+/**
+ * Defines values for chat event.
+ */
+public enum ChatEventKind {
```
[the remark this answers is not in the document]
CHILIU-MSFT: @anuchandy  what do you think? I think ChatEventType looks better.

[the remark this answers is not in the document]
anuchandy: agree with Victor and you, `ChatEventType` is better

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatMessageDeletedEvent.java
CHILIU-MSFT: If we want to use `JacksonSerder` for parsing `JsonObject`, it means we need to put quite a lot extra private properties in event classes. For example, for `ChatMessageReceivedEvent`, we need to add `groupId`, `messageId`, `originalArrivalTime` and `deletetime` with 4 package private set functions. Is there a way to prevent those 4 extra properties to be printed if customer use `JacksonSerder` to print log?

anuchandy: If we want to omit few properties when serializing then Jackson's WRITE_ONLY annotation can be used, like [this](https://github.com/Azure/azure-sdk-for-android/blob/e4a006d662ca2825cf4fe991896c8d3f962574b3/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/models/CommunicationError.java#L29). Btw I'm not sure how important it is to ensure models are printable, at least in java we didn't hear any such ask, users usually access properties using `getters` and same getters used to form string to log.

I assume you mean over the wire  the property is  named`groupId` but in the model we want it to map to a property with the name `threadId`, if that's the case could we use the name attribute of JasonProperty as in, does that solves these categories of issue?

```java
public abstract class BaseEvent {
    @JsonProperty(value = "groupId")
    private String threadId;

   public String getThreadId() {
      return this.threadId;
   }
}
```

CHILIU-MSFT: If there's no requirement for consistency in property names in printed JSON v.s. class field name, above solution 

    @JsonProperty(value = "groupId")
    private String threadId;

is good enough for our scenario. 
Thanks! I'll apply the code change.

anuchandy: Cool, yes, no such requirement; Even in AutoRest, we allow customization to rename getters, where the wire-name is considered more as impl details.

CHILIU-MSFT:
Updated APIView
https://apiview.dev/Assemblies/Review/337c2c89513d482096684c61f1593518

anuchandy (approved):

juancamilor (approved):

PR #777, opened 2021-05-20
Added the ServiceVersion interface to Azure Core.

vcolin7:
Noticed this was missing from Azure Core

### DIFF
--- a/sdk/core/azure-core/src/main/java/com/azure/android/core/util/ServiceVersion.java
+++ b/sdk/core/azure-core/src/main/java/com/azure/android/core/util/ServiceVersion.java
@@ -4,7 +4,7 @@
 package com.azure.android.core.util;
 
 /**
- * A generic interface for sending HTTP requests using the provided service version.
+ * A generic interface for sending service requests using the provided service version.
  */
 public interface ServiceVersion {
     /**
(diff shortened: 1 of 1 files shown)


sdk/core/azure-core/src/main/java/com/azure/android/core/util/ServiceVersion.java
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.android.core.util;
+
+/**
+ * A generic interface for sending HTTP requests using the provided service version.
```
anuchandy, line 7: Maybe we don't have to tie it to HTTP? something like "A generic interface for sending service requests using the provided service version." thoughts?

anuchandy (approved):

sdk/core/azure-core/src/main/java/com/azure/android/core/util/ServiceVersion.java
vcolin7: Good catch, I'll make the change

vcolin7, line 7:
```suggestion
 * A generic interface for sending service requests using the provided service version.
```

PR #779, opened 2021-05-21
Adding support for async stream, pagination types and updating Chat list* APIs to use it.

anuchandy:
Screen Shot 2021-05-21 at 4 10 14 PM


sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/CFBackedPageAsyncStream.java
```diff
@@ -0,0 +1,94 @@
...
+import com.azure.android.core.util.Function;
+import com.azure.android.core.util.Predicate;
+
+import java.util.concurrent.CancellationException;
+
+import java9.util.concurrent.CompletableFuture;
+import java9.util.concurrent.CompletionException;
+
+public final class CFBackedPageAsyncStream<T>
+    implements AsyncStream<PagedResponse<T>> {
```
JianpingChen, line 21: This is just proof of concept, correct? Otherwise, move to com.azure.android.core somewhere.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/ChatImpl.java
```diff
@@ -29,6 +27,8 @@
 import com.azure.android.core.rest.annotation.ServiceMethod;
 import com.azure.android.core.rest.annotation.UnexpectedResponseExceptionType;
 import com.azure.android.core.rest.annotation.UnexpectedResponseExceptionTypes;
+import com.azure.android.core.rest.util.paging.PagedResponse;
+import com.azure.android.core.rest.util.paging.PagedResponseBase;
```
JianpingChen, line 31: Need to modify autorest.android once this is final

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/CFBackedPageAsyncStream.java
anuchandy: It's not proof of concept, the idea is this "base stream" will exist in all `CompletableFuture` based library as a "private" class (hopefully emitted by a hardcoded AutoRest template). This stream will back the core streams returned from public APIs.

sdk/core/azure-core-rest/src/test/java/com/azure/android/core/rest/util/paging/CFBackedPageAsyncStream.java
```diff
@@ -0,0 +1,92 @@
...
+import com.azure.android.core.util.AsyncStream;
+import com.azure.android.core.util.AsyncStreamHandler;
+import com.azure.android.core.util.CancellationToken;
+
+import java.util.concurrent.CancellationException;
+
+import java9.util.concurrent.CompletableFuture;
+import java9.util.concurrent.CompletionException;
+
+final class CFBackedPageAsyncStream<T>
```
JianpingChen, line 18: This is Java. What does "CF" stand for? :-)

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/CFBackedPageAsyncStream.java
anuchandy: None of the core modules depends on any specific async library (CF, Rx, etc..), so if we want to move this to the core, it means we need a new core module for CF with only this class, so it's a trade of btw maintaining a new module vs having a private class in SDKs.

sdk/core/azure-core-rest/src/test/java/com/azure/android/core/rest/util/paging/CFBackedPageAsyncStream.java
anuchandy: :) true. its stands for "CompletableFuture", this is a private class for testing, but the name can be more specific if we prefer so.

sdk/core/azure-core-rest/src/test/java/com/azure/android/core/rest/util/paging/CFStringPageRetriever.java
```diff
@@ -0,0 +1,96 @@
...
+import java.io.UncheckedIOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import java9.util.concurrent.CompletableFuture;
+
+final class CFStringPageRetriever
+    implements Function<String, CompletableFuture<PagedResponse<String>>> {
```
JianpingChen, line 21: I need to understand why "String" is so special. Can this class be generalized to CFPageRetriever<T>?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadAsyncClient.java
```diff
@@ -336,7 +336,7 @@ public String getChatThreadId() {
      * @return the {@link CompletableFuture} that emits list of thread participants in the first page.
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
-    public CompletableFuture<Page<ChatParticipant>> getParticipantsFirstPage() {
+    public CompletableFuture<Page<String, ChatParticipant>> getParticipantsFirstPage() {
```
anuchandy, line 339: Note: Eventually, the public API in SDKs (e.g. Chat) will not return `Page<String, ..>` or `PagedResponse<..>`, see the `listChat()` as an example. The APIs that expose those types in this pr is the one needs to be adapted to new asyncstream / pagedIterable \\cc @CHILIU-MSFT

sdk/core/azure-core-rest/src/test/java/com/azure/android/core/rest/util/paging/CFStringPageRetriever.java
anuchandy: You mean the first param right, its the type of the nextLink (some libraries support non-string nextLink), but for most of the libraries including Chat, we will not force end-user to provide this generic type, they will work with `PagedResponse` that is specialized for `String` nextLink. Just to clarify, this file is a part of the test.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -118,48 +122,38 @@ public ChatThreadAsyncClient getChatThreadClient(String chatThreadId) {
...
-     * @return the {@link CompletableFuture} that emits the first page of chat threads.
+     * @return the paged stream of chat threads of a user.
      */
-    @ServiceMethod(returns = ReturnType.SINGLE)
-    public CompletableFuture<Page<ChatThreadItem>> getChatThreadsFirstPage() {
-        ListChatThreadsOptions listThreadsOptions = new ListChatThreadsOptions();
-        return this.getChatThreadsFirstPage(listThreadsOptions, null)
-            .thenApply(response ->  new PageImpl<>(response.getValue(), response.getContinuationToken()));
+    @ServiceMethod(returns = ReturnType.COLLECTION)
+    public PagedAsyncStream<ChatThreadItem> listChatThreads() {
```
CHILIU-MSFT, line 130: I notice there're 2 functions for listChatThreads
PagedAsyncStream<ChatThreadItem> listChatThreads()
PagedAsyncStream<ChatThreadItem> listChatThreads(ListChatThreadsOptions listThreadsOptions, Context context)

I'll add another 
PagedAsyncStream<ChatThreadItem> listChatThreads(ListChatThreadsOptions listThreadsOptions)

Does it make sense?

vcolin7: Guidelines say we will have an options bag in the maximal overload of a method, including context, so we don't really need to add another overload only for the options :)

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/CFBackedPageAsyncStream.java
```diff
@@ -0,0 +1,94 @@
...
+                    throw logger.logExceptionAsError((RuntimeException) throwable);
+                } else {
+                    // wrap checked exception in a unchecked runtime exception
+                    throw logger.logExceptionAsError(new RuntimeException(throwable));
+                }
+            }
+
+            handler.onNext(pagedResponse);
+
+            final String nextPageId = pagedResponse.getContinuationToken();
```
CHILIU-MSFT, line 81: I see some variables named as "nextPageId" or "pageId" are actually refer to a "nextLink". Is it better to rename them as "nextLink"? Because "nextLink" is not an ID of the page. What do you think?

sdk/communication/azure-communication-chat/src/test/java/com/azure/android/communication/chat/ChatClientTestBase.java
```diff
@@ -174,6 +178,15 @@ public boolean checkParticipantsListContainsParticipantId(List<ChatParticipant>
         return false;
     }
 
+    static void awaitOnLatch(CountDownLatch latch, String method) {
+        long timeoutInSec = 2;
+        try {
+            latch.await(timeoutInSec, TimeUnit.MINUTES);
+        } catch (InterruptedException e) {
+            assertFalse(true, method + " didn't complete within " + timeoutInSec + " minutes");
```
CHILIU-MSFT, line 186: Is the original purpose to use 2 seconds or 2 minutes delay here?
"timeoutInSec" conflicts with "minutes" in error message.

anuchandy: my bad, I meant to name the variable `timeoutInMin`

CHILIU-MSFT:
Updated APIView for this change:
https://apiview.dev/Assemblies/Review/337c2c89513d482096684c61f1593518#

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/CFBackedPageAsyncStream.java
JianpingChen: Thanks for the explanation

JianpingChen (approved):

sdk/communication/azure-communication-chat/src/test/java/com/azure/android/communication/chat/ChatClientTestBase.java
anuchandy: Let me do the test name cleanup in a follow up.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/CFBackedPageAsyncStream.java
anuchandy: This is a good question(I spent a lot of time thinking about it while designing); few overlapping use-cases make it hard to align on a good name. For example, android pagination libraries often use the name page identifier; for some azure services, it can be only a token, not a link (as in URL), and for some services, it's an HTTP/S URL.

collection::getPage(pageId) vs collection::getPage(token) vs collection::getPage(nextLink)

Out of various options, we felt that pageId suite better for the above API, as it is more common to be consumed from android pagination code user writes.

And PagedResponse::continuationToken(), covers nextLink if we treat it as a opaque string. 

But I agree that this is still not aligning perfectly. (Something we also want to observe in user study).

PR #794, opened 2021-06-08
Updated Azure Core's versions after June 2021 beta release.

vcolin7:


sdk/communication/azure-communication-common/build.gradle
```diff
@@ -1,6 +1,6 @@
 ext.publishName = "Microsoft Azure Android Client Library common code For Communication Service"
 description = "This package contains the Android client library common code for Microsoft Azure Communication Service."
-version = "1.1.0-beta1"
+version = "1.1.0-beta.1"
```
JianpingChen, line 3: Might as well make it beta.2. The old name reads funny. It is a mistake that should be fixed anyway

sdk/communication/azure-communication-chat/build.gradle
```diff
@@ -34,24 +34,24 @@ static def getBuildVariant() {
```
JianpingChen, line 34: Update Chat version

JianpingChen:
Update ChangeLog.md files

sdk/communication/azure-communication-common/build.gradle
JianpingChen: Also dependency on Core changed.

vcolin7: Common has not been published yet with `beta.1` after GA, I think we should not skip that version.

sdk/communication/azure-communication-chat/build.gradle
vcolin7: This PR is not for preparing Chat for the GA release. I think we should update the version in a separate PR.

sdk/communication/azure-communication-common/build.gradle
vcolin7: What do you mean on the second comment? I don't understand if that means we should have different dependencies or if the version changed. Is it something else?

JianpingChen: Sorry about the confusion. I was thinking the published version should have been named "beta.1". Now we are releasing a new version might as well name it "beta.2". One of the reason is this new version is different from (beta1 aka beta.1) in that it depends on new version of android core.

vcolin7: But there is no published version `1.1.0-beta.1`. We should not jump to `1.1.0-beta.2` if `1.1.0-beta.1` has never been released.

JianpingChen:
> Update ChangeLog.md files

Also ReadMe.md files. I always forgot

sdk/communication/azure-communication-chat/build.gradle
JianpingChen: Sorry, I thought it was. If we are expecting another PR for releasing, sure. Just ignore me.

sdk/communication/azure-communication-common/build.gradle
anuchandy: We published GA version `1.0.0` in the last release after a series of beta versions (`1.0.0-beta.1` to `1.0.0-beta.8`). The GA version `1.0.0` uses new azure-core modules. The `build.gradle` in "the repro" should version the next "dev-candidate" as `1.1.0-beta.1`. (1.0.0 => 1.1.0-beta.1), which is what we'rere doing in this pr.

anuchandy (approved):

PR #796, opened 2021-06-09
Fix issue in APIView feedback

CHILIU-MSFT:


JianpingChen (approved):

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/AddChatParticipantsResult.java
```diff
@@ -4,12 +4,12 @@
 
 package com.azure.android.communication.chat.models;
 
-import com.azure.android.core.rest.annotation.Immutable;
+import com.azure.android.core.rest.annotation.Fluent;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.List;
 
 /** Result of the add chat participants operation. */
-@Immutable
```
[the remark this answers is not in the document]
CHILIU-MSFT: There's a setInvalidParticipants() in this class.  Srikanta mentioned there should not be setter for Immutable class

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClientBuilder.java
```diff
@@ -248,15 +218,17 @@ public ChatAsyncClient buildAsyncClient() {
...
 
     private HttpPipeline createHttpPipeline(HttpClient httpClient,
                                             HttpPipelinePolicy authorizationPolicy,
                                             List<HttpPipelinePolicy> additionalPolicies) {
 
         List<HttpPipelinePolicy> policies = new ArrayList<HttpPipelinePolicy>();
-        policies.add(authorizationPolicy);
+        if (authorizationPolicy != null) {
+            policies.add(authorizationPolicy);
+        }
```
JianpingChen, line 231: authorizationPolicy should be mandatory, should it not?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadClientBuilder.java
```diff
@@ -114,6 +100,20 @@ public ChatThreadClientBuilder addPolicy(HttpPipelinePolicy pipelinePolicy) {
...
     }
 
+    /**
+     * Sets the {@link RetryPolicy} that will attempt to retry failed requests, if applicable.
+     *
+     * @param retryPolicy the retryPolicy value.
+     * @return the AzureCommunicationChatServiceImplBuilder.
+     */
+    public ChatThreadClientBuilder retryPolicy(RetryPolicy retryPolicy) {
+        if (retryPolicy == null) {
```
JianpingChen, line 110: Objects.RequireNonNull

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadClientBuilder.java
```diff
@@ -256,7 +248,9 @@ private HttpPipeline createHttpPipeline(HttpClient httpClient,
                                             List<HttpPipelinePolicy> additionalPolicies) {
 
         List<HttpPipelinePolicy> policies = new ArrayList<HttpPipelinePolicy>();
-        policies.add(authorizationPolicy);
+        if (authorizationPolicy != null) {
+            policies.add(authorizationPolicy);
+        }
```
JianpingChen, line 253: How can authorizationPolicy be optional here?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClientBuilder.java
CHILIU-MSFT: Good question. This is the part I want to be reviewed.

There're 2 ways for customer to set authorizationPolicy: 1) add a HttpPipelinePolicy for it, 2) add a CommunicationTokenCredential and inside this client builder we create this authorizationPolicy from the credential.

In API review,  Srikanta suggested to remove function credentialPolicy() from client builder because we can use addPolicy() for it. If customer is using addPolicy() to set a HttpPipelinePolicy, it would be added to this.customPolicies. If customer is setting CommunicationTokenCredential, we would create this authorizationPolicy for him. So, authorizationPolicy would only present when customer is setting CommunicationTokenCredential. That's why it can be optional.

Besides, this means we lost the ability to validate whether customer has set a HttpPipelinePolicy or CommunicationTokenCredential for authorization. Would this be an issue?

CHILIU-MSFT: Same for ChatThreadClientBuilder

vcolin7: We should validate that there is some mechanism for authorization or else we throw. I'm not a fan of using the generic `addPolicy()` to add the authorization policy, as it can make it hard to discover you need to provide an instance of `BearerTokenAuthenticationPolicy`. If we want to go this rout, we should be extremely clear in the JavaDoc about needing to authenticate the clients somehow and what options customers have. Builders should also throw when there's no form of authenticating when calling `buildClient()`/`buildAsyncClient()` and be super clear about what the problem is in the error message.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -244,8 +246,10 @@ public ChatThreadAsyncClient getChatThreadClient(String chatThreadId) {
 
     /**
      * Receive real-time messages and notifications.
+     * @param skypeUserToken the skype user token
+     * @param context the app context of the app
      */
-    public void startRealtimeNotifications() {
+    public void startRealtimeNotifications(String skypeUserToken, Context context) {
```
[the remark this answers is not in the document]
CHILIU-MSFT: No. SignalingClient is a final private property which initialized in constructor of ChatAsyncClient.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClientBuilder.java
JianpingChen: I agree with Victor here. Chat client has to be clear on what kind of credential it supports. If we only support TokenCredential, we make that clear by throwing in the builder when it is null. We cannot make the promise that any kind of authentication policy would work. Therefore, assuming customer should somehow set an authentication policy is just not a safe assumption.

CHILIU-MSFT:
Add @throw annotations
Make communicationTokenCredential mandatory in builder
Put authorizationPolicy after retryPolicy in builder

sdk/communication/azure-communication-chat/src/test/java/android/util/Base64.java
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package android.util;
+
```
JianpingChen, line 5: Interesting that Java allows you to hijack someone else's code. :-)

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -67,6 +71,8 @@ public ChatThreadAsyncClient getChatThreadClient(String chatThreadId) {
      * Creates a chat thread.
      *
      * @param options Options for creating a chat thread.
+     * @throws ChatErrorResponseException thrown if the request is rejected by server.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
```
vcolin7, line 75: I don't think we need to add `RequestException` to the JavaDoc. What do you think @srnagar, @JonathanGiles?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -67,6 +71,8 @@ public ChatThreadAsyncClient getChatThreadClient(String chatThreadId) {
      * Creates a chat thread.
      *
      * @param options Options for creating a chat thread.
+     * @throws ChatErrorResponseException thrown if the request is rejected by server.
```
vcolin7, line 74: nit; Using the word `thrown` reads weird after seeing `@throws` right before. This applies to all instances of this in the project's JavaDoc.
```suggestion
     * @throws ChatErrorResponseException if the request is rejected by server.
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -243,81 +256,52 @@ public ChatThreadAsyncClient getChatThreadClient(String chatThreadId) {
     }
 
     /**
-     * Receive real-time messages and notifications.
+     * Receive real-time notifications.
+     * @param skypeUserToken the skype user token
+     * @param context the app context of the app
```
vcolin7, line 261: What about using something like this? It's a bit more descriptive and avoids using the word 'app' twice.
```suggestion
     * @param context the Android app context
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClientBuilder.java
```diff
@@ -118,10 +102,26 @@ public ChatClientBuilder addPolicy(HttpPipelinePolicy pipelinePolicy) {
         return this;
     }
 
+    /**
+     * Sets the {@link RetryPolicy} that will attempt to retry failed requests, if applicable.
+     *
+     * @param retryPolicy the retryPolicy value.
+     * @throws NullPointerException thrown if retryPolicy is null.
+     * @return the AzureCommunicationChatServiceImplBuilder.
```
vcolin7, line 110:
```suggestion
     * @return The updated {@link ChatClientBuilder} object.
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadClientBuilder.java
```diff
@@ -114,10 +104,26 @@ public ChatThreadClientBuilder addPolicy(HttpPipelinePolicy pipelinePolicy) {
         return this;
     }
 
+    /**
+     * Sets the {@link RetryPolicy} that will attempt to retry failed requests, if applicable.
+     *
+     * @param retryPolicy the retryPolicy value.
+     * @throws NullPointerException thrown if retryPolicy is null.
+     * @return the AzureCommunicationChatServiceImplBuilder.
```
vcolin7, line 112:
```suggestion
     * @return The updated {@link ChatClientBuilder} object.
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
CHILIU-MSFT: I notice in JAVA SDK, RequestException is added. @vcolin7

vcolin7: Oh my bad, I meant to say `RuntimeException`.

CHILIU-MSFT: I also meant to say in JAVA SDK RuntimeException is added. I made the same mistake...

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -44,16 +47,17 @@
     private final SignalingClient signalingClient;
     private final ChatImpl chatClient;
 
-    ChatAsyncClient(AzureCommunicationChatServiceImpl chatServiceClient, SignalingClient signalingClient) {
+    ChatAsyncClient(AzureCommunicationChatServiceImpl chatServiceClient) {
         this.chatServiceClient = chatServiceClient;
-        this.signalingClient = signalingClient;
+        this.signalingClient = new CommunicationSignalingClient();
```
anuchandy, line 52: If I understand correctly - with the `new CommunicationSignalingClient()`, we ALWAYS loads the `TrouterClient` into JVM, earlier this was done only if the user-provided the notification params in the builder. I think the new aggressive instantiation will be a problem if the user wants to use "chat SDK without notification" + "calling SDK" in the same app, given loading two trouter is an issue at least now (that router team is working on).

CHILIU-MSFT: You're right. Maybe we could update to create signalingClient instance when user calls startRealtimeNotification the first time?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadClientBuilder.java
```diff
@@ -114,10 +104,26 @@ public ChatThreadClientBuilder addPolicy(HttpPipelinePolicy pipelinePolicy) {
...
 
+    /**
+     * Sets the {@link RetryPolicy} that will attempt to retry failed requests, if applicable.
+     *
+     * @param retryPolicy the retryPolicy value.
+     * @throws NullPointerException if retryPolicy is null.
+     * @return The updated {@link ChatThreadClientBuilder} object.
+     */
+    public ChatThreadClientBuilder retryPolicy(RetryPolicy retryPolicy) {
+        if (retryPolicy == null) {
```
vcolin7, line 115: `retryPolicy` can be `null` if we have a default value to set it to when not present, which we do below in `applyRequiredPolicies()`

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadClientBuilder.java
```diff
@@ -114,10 +104,26 @@ public ChatThreadClientBuilder addPolicy(HttpPipelinePolicy pipelinePolicy) {
...
+    }
+
     /**
      * Sets the {@link HttpLogOptions} for service requests.
      *
      * @param logOptions The logging configuration to use when sending and receiving HTTP requests/responses.
+     * @throws NullPointerException if logOptions is null.
      * @return The updated {@link ChatThreadClientBuilder} object
      */
     public ChatThreadClientBuilder httpLogOptions(HttpLogOptions logOptions) {
```
vcolin7, line 129: `logOptions` can be `null` if we have a default value to set it to when not present, [which we do](https://github.com/Azure/azure-sdk-for-android/blob/209ef5b97c5f2bb7a127db07febeee68bec1a0e6/sdk/core/azure-core-http/src/main/java/com/azure/android/core/http/policy/HttpLoggingPolicy.java#L60).

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadClientBuilder.java
CHILIU-MSFT: Yes. It can be null. If user does not call this function, it would use the default retry policy. Here we only want to check when user calls retryPolicy(RetryPolicy retryPolicy), he should provide a non-null retryPolicy.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatThreadClientBuilder.java
CHILIU-MSFT: Same as retry policy. Here we only want to check when user calls httpLogOptions(HttpLogOptions logOptions), he should provide a non-null logOptions.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
CHILIU-MSFT: Verified. With current implementation "the initialization of trouter is moved to start() method, so types should be loading only when start() is called", if I put chat & calling in the same App with excluding trouter from chat, the App could start and chat functionality works fine if you don't call startRealtimeNotifications().
We are good here.

JianpingChen (approved):

anuchandy (approved):

PR #801, opened 2021-06-16
Update Chat & Common version after release

CHILIU-MSFT:


vcolin7 (commented):
LGTM so far. Is there any other place we should update these versions @JianpingChen?

JianpingChen (approved):

JianpingChen:
> LGTM so far. Is there any other place we should update these versions @JianpingChen?

I think it is OK for Android. For Java we have to take care of code coverage and version_client.txt but I couldn't find equivalents

PR #826, opened 2021-07-23
Added Typing Notification Options for Sender Display Name

angiurgiu:
Added Typing Notification Options for Sender Display Name
- updated swagger
- added additional client methods taking the options
- added additional tests, updated and ran the E2E main activity test suite
- updated changelog


samples/sample-chat-app/src/main/java/com/azure/android/communication/chat/sampleapp/MainActivity.java
```diff
@@ -392,7 +393,10 @@ public void sendTypingNotification(View view) {
 
         if (chatThreadAsyncClient != null) {
             try {
-                chatThreadAsyncClient.sendTypingNotification().get();
+                TypingNotificationOptions options = new TypingNotificationOptions();
+                options.setSenderDisplayName("Sender Display Name");
+
+                chatThreadAsyncClient.sendTypingNotification(options).get();
```
JianpingChen, line 399: The get call reads funny here since no variable is receiving the get result. Please comment if there is a good reason.

JianpingChen (approved):

samples/sample-chat-app/src/main/java/com/azure/android/communication/chat/sampleapp/MainActivity.java
angiurgiu: The existing test calling the other method version does this already, and since I don't know much about CompletableFuture in Android I did the same. Do you this can be safely removed without altering the test logic?

angiurgiu: All the tests in this class always do .get(), even with no variables to save the result. So I'd be inclined to keep the same pattern

PR #841, opened 2021-08-13
Added support FCM push notifications

CHILIU-MSFT:
Added support FCM push notifications
Refactoring real time notification interface to remove skypetoken


sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -275,7 +282,65 @@ public void stopRealtimeNotifications() {
...
+    /**
+     * Add handler for a chat event for push notifications.
+     * @param chatEventType the chat event type
+     * @param listener the listener callback function
+     * @throws IllegalStateException if push notifications has not started yet.
+     */
+    public void addPushNotificationHandler(ChatEventType chatEventType, PushNotificationCallback listener) {
+        if (!this.pushNotificationClient.hasStarted()) {
+            throw logger.logExceptionAsError(new IllegalStateException(
+                "You must call startPushNotifications before you can subscribe to push notifications."
```
vcolin7, line 326:
```suggestion
                "You must call startPushNotifications(String) before you can subscribe to push notifications."
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -291,7 +356,7 @@ public void addEventHandler(ChatEventType chatEventType, RealTimeNotificationCal
     }
 
     /**
-     * Remove handler from a chat event.
+     * Remove handler from a chat event for real-time notifications.
```
vcolin7, line 359: nit; Let's agree to use one of `real time`, `real-time` or `realtime`. I think Firebase uses `realtime` from what I saw online.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClient.java
```diff
@@ -177,23 +179,72 @@ public void deleteChatThread(String chatThreadId) {
...
     /**
      * Receive real-time notifications.
-     * @param skypeUserToken the skype user token
      * @param context the Android app context
      * @throws RuntimeException if real-time notifications failed to start.
      */
-    public void startRealtimeNotifications(String skypeUserToken, Context context) {
-        this.client.startRealtimeNotifications(skypeUserToken, context);
+    public void startRealtimeNotifications(Context context) {
+        this.client.startRealtimeNotifications(context);
```
vcolin7, line 186: Same comment from `ChatAsyncClient` about this being a breaking change.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClient.java
```diff
@@ -177,23 +179,72 @@ public void deleteChatThread(String chatThreadId) {
...
     }
 
     /**
      * Stop receiving real-time notifications.
+     * All registered handlers would be removed.
```
vcolin7, line 191:
```suggestion
     * All registered handlers will be removed.
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -275,7 +282,65 @@ public void stopRealtimeNotifications() {
...
+        if (this.pushNotificationClient.hasStarted()) {
+            return;
+        }
+
+        this.pushNotificationClient.startPushNotifications(deviceRegistrationToken);
+    }
+
+    /**
+     * Unregister current device from receiving incoming push notifications.
+     * All registered handlers would be removed.
```
vcolin7, line 299:
```suggestion
     * All registered handlers will be removed.
```

sdk/communication/azure-communication-chat/CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
...
+        - Added `removePushNotificationHandler`
+    - ChatClient:
+        - Added `startPushNotifications`
+        - Added `stopPushNotifications`
+        - Added `handlePushNotification`
+        - Added `addPushNotificationHandler`
+        - Added `removePushNotificationHandler`
+     - Added `ChatPushNotification` as the type for passing push notification payload
+     - Added `PushNotificationCallback` as the type for customized push notification handler
+     - Added instruction in README.md in sample app for steps to setup FCM push notification
```
vcolin7, line 19: nit;
```suggestion
     - Added instructions in README.md in sample app for steps to setup FCM push notifications
```

sdk/communication/azure-communication-chat/env-test.properties
```diff
@@ -1,9 +1,14 @@
 trouter.client.version="99116/0.0.0.0/"
 trouter.max.registration.ttls="3600"
-platform="SPOOL"
+platform="Android"
 platform.ui.version="0.0.0.0"
 
 trouter.hostname="go.trouter-int.skype.net"
 trouter.registration.hostname.and.base.path="edge.skype.net/registrar/testenv"
 trouter.application.id="AcsAndroid_test"
 trouter.template.key="AcsAndroid_Chat_test_1.2"
```
vcolin7, line 9: Any reason why the above is `1.3` and this is `1.2`?

sdk/communication/azure-communication-chat/env-test.properties
```diff
@@ -1,9 +1,14 @@
...
+
+pushnotification.registrar.service.url="https://edge.skype.net/registrar/testenv/v2/registrations"
+pushnotification.application.id="AcsAndroid"
+pushnotification.template.key="AcsAndroid.AcsNotify_Chat_1.0"
```
vcolin7, line 13:
```suggestion
pushnotification.application.id="AcsAndroid_test"
pushnotification.template.key="AcsAndroid_test.AcsNotify_Chat_1.0"
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        this.isPushNotificationsStarted = true;
+        this.pushNotificationListeners.clear();
+        this.logger.info(" Registered push notification successfully!");
+    }
+
+    /**
+     * Unregister the current device from receiving incoming push notifications.
+     * All registered handlers would be removed.
```
vcolin7, line 79:
```suggestion
     * All registered handlers will be removed.
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrarClient.java
```diff
@@ -0,0 +1,212 @@
...
+
+        @JsonProperty(value = "registrationId")
+        private String registrationId;
+
+        @JsonProperty(value = "transports")
+        private Transports transports;
+    }
+
+    RegistrarClient() {
+        this.httpClient = new OkHttpAsyncHttpClientBuilder().build();
```
vcolin7, line 101: We probably should use [this](https://github.com/Azure/azure-sdk-for-android/blob/71214038c6d7ba05b3d2a4b11ceec65771107ef5/sdk/core/azure-core-http/src/main/java/com/azure/android/core/http/HttpClient.java#L42) to keep the HTTP client library pluggable:
```suggestion
        this.httpClient = HttpClient.createDefault();
```

vcolin7 (changes requested):
Looks good for the most part. I left some comments I'd like to be addressed before approving though.

sdk/communication/azure-communication-chat/env-test.properties
CHILIU-MSFT: I'll fix this.

samples/sample-chat-app/README.md
```diff
@@ -6,13 +6,20 @@ Setup:
...
+5.3 Follow this link https://docs.microsoft.com/azure/communication-services/concepts/notifications#notification-hub-provisioning to link your ACS resource with your ACS resource.
+6. Run "sample-chat-app" from the IDE by clicking the run button on the top right corner.
 
 Functions:
-* Use the button 'Start real time notifications', to start the real time notifications. (Implementation pending)
-* Use the button 'Register real time notifications listener' to register a listener for 'chatMessageReceived' events. (Implementation pending)
-* Use the button 'Unregister real time notifications listener' to unregister a listener for 'chatMessageReceived' events. (Implementation pending)
+* Use the button 'Start real time notifications', to start the real time notifications.
+* Use the button 'Register real time notifications listener' to register listeners for all events.
+* Use the button 'Unregister real time notifications listener' to unregister a listener for 'chatMessageReceived' events.
```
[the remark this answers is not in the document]
CHILIU-MSFT: Yes. It should unregister all events. But in current sample app code, it does only unregister chatMessageReceived event. I'll fix it.

samples/sample-chat-app/README.md
```diff
@@ -6,13 +6,20 @@ Setup:
...
 
-> After registering to the event, you need to use postman or JS client to create a thread with the same token you used in the test app, then send a message to the thread. Observe log and app UI to see if the test app receives real time notifications.
+* Use the button 'Start push notifications', to start FCM push notifications.
+* Use the button 'Register push notifications listener' to register listeners for all FCM push notifications events.
```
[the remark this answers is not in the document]
CHILIU-MSFT: Yes. We provide a 
public public void removePushNotificationHandler(ChatEventType chatEventType, PushNotificationCallback listener)
 in both ChatAsyncClient and ChatClient to unregister listeners.

CHILIU-MSFT:
https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/21694/Android-Details

This is a wiki for the call flows behind.

sdk/communication/azure-communication-chat/env-test.properties
CHILIU-MSFT: In INT PNH template for fcm push notification we don't put a "test" in appId/templateKey, which is a little inconsistency with what we have for trouter.

https://skype.visualstudio.com/SCC/_git/async_notification_templates?path=%2FTestEnv%2Ffcm%2FACSNotify%2FChat%2F1.0%2Fbase.json

We use this appId/templateKey because Calling does the same: using same appId/templateKey in both INT & PROD for FCM. We are making this part consist with Calling. https://skype.visualstudio.com/SCC/_git/async_notification_templates?path=%2FTestEnv%2Ffcm%2FACSNotify%2FCalling%2F2.0%2Fbase.json

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatUserEvent.java
```diff
@@ -21,6 +21,12 @@
      * The Id of the event recipient. A string property in notification payload.
```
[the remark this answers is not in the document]
CHILIU-MSFT: They are actually refer to the same ID.
In old Event payload, there's a property named as "recipientId" with a value of MRI without 8: prefix, which we identified as a BUG. To prevent break existing code, we added a property "recipientMri" for full MRI. 
So, here we did something special for "recipientId" & "recipientMri". In line 73 - 81 of this file, if recipientMri exists, we take that to generate the communicationIdentifier, otherwise recipientId.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatPushNotification.java
```diff
@@ -0,0 +1,40 @@
...
+import com.azure.android.core.rest.annotation.Fluent;
+
+import java.io.Serializable;
+import java.util.Map;
+
+/**
+ * The chat push notification payload from FCM.
+ */
+@Fluent
+public final class ChatPushNotification implements Serializable {
```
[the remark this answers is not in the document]
CHILIU-MSFT: In current sample app implementation, in line 43 of MyFirebaseMessagingService.java we add this object into the Intent.
Then, in line 113 of MainActivity.java, we get ChatPushNotification object from the Intent and pass it to chatAsyncClient to let it handled by registered handler. With Intent, this class needs to be Serializable. 

Do you have any idea whether there could be any alternative implementation?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -255,16 +263,15 @@ public ChatThreadAsyncClient getChatThreadClient(String chatThreadId) {
...
      */
-    public void startRealtimeNotifications(String skypeUserToken, Context context) {
+    public void startRealtimeNotifications(Context context) {
         if (this.signalingClient.hasStarted()) {
             return;
         }
 
-        this.signalingClient.start(skypeUserToken, context);
+        this.signalingClient.start(context);
     }
```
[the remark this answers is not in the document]
CHILIU-MSFT: The idea here is to store the credential instead of asking for skypeUserToken every time.

For Push Notification, for both start/stop, a valid skypeUserToken is required. So I updated the ChatAsyncClient to store the credential and request for token when it's required. With this update, ChatAsyncClient would expose following functions for realtime/push notification:
public void startRealtimeNotifications(Context context)
public void stopRealtimeNotifications()
public void startPushNotifications(String deviceRegistrationToken)
public void stopPushNotifications()

But since this is a breaking change and we can't do it for stable release. I think I would revert the interface change for realtime notification and expose the following:
public void startRealtimeNotifications(String skypeUserToken, Context context)
public void stopRealtimeNotifications()
public void startPushNotifications(String skypeUserToken, String deviceRegistrationToken)
public void stopPushNotifications(String skypeUserToken)

What do you think?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatPushNotification.java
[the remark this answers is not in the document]
vcolin7: I think [`Parceable`](https://developer.android.com/reference/android/os/Parcelable) is more idiomatic to Android, plus it provides better performance [when it comes to passing many objects](https://stackoverflow.com/a/22350463/4967034). It takes a bit more code to implement the `Parceable` interface when compared to `Serializable`, but I think [if you are not looking to persist the object](https://stackoverflow.com/questions/3323074/android-difference-between-parcelable-and-serializable#comment77665944_23647471), then we should go with solutions that were made with mobile in mind.

Here's a good [comparison](https://stackoverflow.com/a/23647471/4967034) between the code needed for both.

@anuchandy @JonathanGiles any thoughts on this?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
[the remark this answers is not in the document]
vcolin7: If you think users won't be passing a `skypeUserToken` in the future, you could add overloads that don't include said parameter instead. //cc @JonathanGiles in case he thinks we should do something else here.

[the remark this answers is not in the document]
CHILIU-MSFT: Do you mean leave 
public void startRealtimeNotifications(String skypeUserToken, Context context)
there and add an extra
public void startRealtimeNotifications(Context context)
?

sdk/communication/azure-communication-chat/env-test.properties
vcolin7: I can't access the repositories you shared but if you guys are planning on making a change to be consistent with Calling then that's fine by me.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/models/ChatPushNotification.java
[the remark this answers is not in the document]
JonathanGiles: Generally I see implementation of the `Serializable` interface as a code smell. I would prefer we didn't do this unless there was a real need for it.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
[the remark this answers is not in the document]
JonathanGiles: I forget, has this library already GA'd? If so, we can't remove existing API, and the code in your previous comment is accurate.. If it has not yet GA'd, then it is fine to remove it.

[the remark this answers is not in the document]
CHILIU-MSFT: Got it. Yes, this API is GA'd. I'll add an overload for it instead of update it. Thanks! @JonathanGiles

sdk/communication/azure-communication-chat/CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
...
+        - Added `stopPushNotifications`
+        - Added `handlePushNotification`
+        - Added `addPushNotificationHandler`
+        - Added `removePushNotificationHandler`
+     - Added `ChatPushNotification` as the type for passing push notification payload
+     - Added `PushNotificationCallback` as the type for customized push notification handler
+     - Added instructions in README.md in sample app for steps to setup FCM push notifications
+     - Added code in sample app for FCM push notification end to end scenario
+- Added `startRealtimeNotifications(Context context)` to ChatAsyncClient
+- Added `startRealtimeNotifications(Context context)` to ChatClient
```
vcolin7, line 22: We are using a different header now for new features. Also, you might want to make it clear these are methods users can call.
```suggestion
### Features Added
- Added support FCM push notifications:
    - Added methods to `ChatAsyncClient`:
        - `startPushNotifications`
        - `stopPushNotifications`
        - `handlePushNotification`
        - `addPushNotificationHandler`
        - `removePushNotificationHandler`
    - Added methods to `ChatClient`:
        - `startPushNotifications`
        - `stopPushNotifications`
        - `handlePushNotification`
        - `addPushNotificationHandler`
        - `removePushNotificationHandler`
     - Added `ChatPushNotification` as the type for passing a push notification payload.
     - Added `PushNotificationCallback` as the type for a customized push notification handler.
     - Added instructions in README.md in sample app for steps to setup FCM push notifications.
     - Added code in sample app for FCM push notification end to end scenario.
- Added `startRealtimeNotifications(Context context)` to `ChatAsyncClient`.
- Added `startRealtimeNotifications(Context context)` to `ChatClient`.
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClientBuilder.java
```diff
@@ -251,7 +251,7 @@ private HttpPipeline createHttpPipeline(HttpClient httpClient,
     }
 
     private void applyRequiredPolicies(List<HttpPipelinePolicy> policies, HttpPipelinePolicy authorizationPolicy) {
-        policies.add(new UserAgentPolicy(null, "azure-communication-chat", "1.0.0"));
+        policies.add(new UserAgentPolicy(null, "azure-communication-chat", "1.1.0-beta.2"));
```
vcolin7, line 254: We should avoid using hard-coded strings for things like these. I'd recommend introducing a constant or something of the sort.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -0,0 +1,194 @@
...
+        String skypeUserToken;
+        try {
+            skypeUserToken = communicationTokenCredential.getToken().get().getToken();
+        } catch (InterruptedException e) {
+            throw logger.logExceptionAsError(
+                new RuntimeException("Get skype user token failed for push notification: " + e.getMessage()));
+        } catch (ExecutionException e) {
+            throw logger.logExceptionAsError(
+                new RuntimeException("Get skype user token failed for push notification: " + e.getMessage()));
+        }
```
vcolin7, line 66: We can collapse these catch blocks.
```suggestion
        } catch (ExecutionException | InterruptedException e) {
            throw logger.logExceptionAsError(
                new RuntimeException("Get skype user token failed for push notification: " + e.getMessage()));
        }
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/signaling/package-info.java
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for
+// license information.
+//
+
+/** Package containing the classes for Signaling. Realtime notification service. */
```
vcolin7, line 6: nit; Formatting.
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License. See License.txt in the project root for license information.

/** Package containing the classes for the Signaling Realtime notification service. */
```

samples/sample-chat-app/src/main/java/com/azure/android/communication/chat/sampleapp/MainActivity.java
```diff
@@ -82,12 +93,27 @@
...
+        public void onReceive(Context context, Intent intent) {
+            ChatPushNotification pushNotification =
+                (ChatPushNotification) intent.getParcelableExtra("PushNotificationPayload");
+
+            Log.d(TAG, "Push Notification received in MainActivity: " + pushNotification.getPayload());
+
+            if (chatAsyncClient != null) {
+                Log.d(TAG, "Passing push notification to chatAsyncClient.");
+                chatAsyncClient.handlePushNotification(pushNotification);
+            }
```
vcolin7, line 114: Is there any scenario where the client could be `null` but this object isn't?

vcolin7 (commented):
Great work on this @CHILIU-MSFT. Added a few more comments :)

samples/sample-chat-app/src/main/java/com/azure/android/communication/chat/sampleapp/MainActivity.java
CHILIU-MSFT: Actually, no. We would call "createChatAsyncClient()" before register the receiver in "onCreate". I think I can remove this check.

vcolin7 (approved):

PR #861, opened 2021-09-03
Changed UserAgentPolicy to correctly follow the guidelines about handling an `applicationId`.

vcolin7:
The guidelines say an `applicationId` cannot be longer than 24 characters and that it cannot contain whitespaces. We are aligning with Java in this regard and throwing when a value that does not meet the two aforementioned criteria is provided.


CHILIU-MSFT (approved):
LGTM

JonathanGiles (changes requested):
Don't add the CoreUtils as public API (and realistically, for the API you've added, it doesn't need to exist yet).

anuchandy (approved):

JonathanGiles (approved):

PR #864, opened 2021-09-08
Update ACS Chat SDK swagger to 2021-09-07 version

CHILIU-MSFT:
Update ACS Chat SDK swagger to 2021-09-07 version


sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatServiceVersion.java
```diff
@@ -10,7 +10,7 @@
  */
 public enum ChatServiceVersion implements ServiceVersion {
     V2021_03_07("2021-03-07"),
```
CHILIU-MSFT, line 12: @LuChen-Microsoft  @juancamilor  Should we leave the stable version here or we should remove it?

vcolin7: That would be a breaking change, so we have to keep it :)

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/AzureCommunicationChatServiceImplBuilder.java
```diff
@@ -154,7 +154,7 @@ public AzureCommunicationChatServiceImplBuilder addPolicy(HttpPipelinePolicy cus
      */
     public AzureCommunicationChatServiceImpl buildClient() {
         if (apiVersion == null) {
-            this.apiVersion = "2021-04-05-preview6";
+            this.apiVersion = "2021-09-07";
```
vcolin7, line 157: Is this class auto-generated? Else we can use `ChatServiceVersion` to avoid updating this manually after every service version update.

vcolin7 (approved):
LGTM, added a couple comments.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/AzureCommunicationChatServiceImplBuilder.java
CHILIU-MSFT: Yes. This class is auto-generated.

PR #895, opened 2021-10-02
Add Kotlin as dependency to sampleapp module

DeFuex:
Adding Kotlin as dependency to the sampleapp module. :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext.kotlin_version = '1.6.0-M1'
     repositories {
         google()
         mavenCentral()
@@ -37,6 +38,7 @@ buildscript {
         classpath ("com.vanniktech:gradle-android-junit-jacoco-plugin:$gradleJunitJacocoPluginVersion")
         classpath ("gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:$gradleSpotbugsPluginVersion")
         classpath ("com.google.gms:google-services:$googleGmsServicesPluginVersion")
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
 
--- a/samples/sample-chat-app/build.gradle
+++ b/samples/sample-chat-app/build.gradle
@@ -2,6 +2,7 @@ plugins {
     id 'com.android.application'
     id 'com.google.gms.google-services'
 }
+apply plugin: 'kotlin-android'
 
 android {
     compileSdkVersion 30
@@ -46,4 +47,9 @@ dependencies {
     implementation platform('com.google.firebase:firebase-bom:28.1.0')
     implementation 'com.google.firebase:firebase-analytics'
     implementation 'com.google.firebase:firebase-messaging'
+    implementation "androidx.core:core-ktx:1.6.0"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+}
+repositories {
+    mavenCentral()
 }


build.gradle
```diff
@@ -1,6 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext.kotlin_version = '1.6.0-M1'
```
vcolin7, line 4: `1.6.0-M1` is [an early access version](https://kotlinlang.org/docs/eap.html#build-details) of the Kotlin plugin. We use stable version for all our tooling, so I'd suggest using version `1.5.31`.

vcolin7: That said, let's not add a new line to add this to `ext` when can simply add to the array below.
```suggestion
```

vcolin7: Additionally, we use `camelCase` for naming all our variables, so this should end up looking like the following in the `ext` array below:

```groovy
        kotlinGradlePluginVersion = "1.5.31"
```

build.gradle
```diff
@@ -37,6 +38,7 @@ buildscript {
         classpath ("com.vanniktech:gradle-android-junit-jacoco-plugin:$gradleJunitJacocoPluginVersion")
         classpath ("gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:$gradleSpotbugsPluginVersion")
         classpath ("com.google.gms:google-services:$googleGmsServicesPluginVersion")
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
vcolin7, line 41: nit;
```suggestion
        classpath ("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version")
```

samples/sample-chat-app/build.gradle
```diff
@@ -2,6 +2,7 @@ plugins {
     id 'com.android.application'
     id 'com.google.gms.google-services'
 }
+apply plugin: 'kotlin-android'
```
vcolin7, line 5: nit;
```suggestion

apply plugin: "kotlin-android"
```

samples/sample-chat-app/build.gradle
```diff
@@ -46,4 +47,9 @@ dependencies {
     implementation platform('com.google.firebase:firebase-bom:28.1.0')
     implementation 'com.google.firebase:firebase-analytics'
     implementation 'com.google.firebase:firebase-messaging'
+    implementation "androidx.core:core-ktx:1.6.0"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+}
+repositories {
+    mavenCentral()
```
vcolin7, line 54: We have already added `mavenCentral()` to our list of repositories for all projects in the project root's `build.gradle` file.
```suggestion
```

build.gradle
vcolin7: Is there any reason we want to add this here if we only want to use it for the Sample Chat App?

vcolin7 (commented):
Hi @DeFuex! Thanks a lot for your contribution. I have reviewed it and added some comments. As I am not a Kotlin expert, I have a question as well: what is the reason we might want to add the Kotlin plugin to the Sample Chat App's `build.gradle`? Is it so the project can also use Kotlin in the source code instead of only Java?

build.gradle
DeFuex: > Hi @DeFuex! Thanks a lot for your contribution. I have reviewed it and added some comments. As I am not a Kotlin expert, I have a question as well: what is the reason we might want to add the Kotlin plugin to the Sample Chat App's `build.gradle`? Is it so the project can also use Kotlin in the source code instead of only Java?
> 
> I ask this because we are not looking to support Kotlin in the Azure SDK for Android for the moment. You can use our libraries with Kotlin applications, but Java is the SDK's programming language of choice.

Hi :), before i change anything here according to the review (thanks by the way), there's a few things to consider trying out Kotlin at least once https://kotlinlang.org/docs/comparison-to-java.html#what-java-has-that-kotlin-does-not . I'm not saying you ultimately should use it if you are fine with Java, but for what i've seen in the recent projects i was working with it, it reduces quite some boilerplate or makes the flow for the development easier in general. It leans on Java, but it offers a concise set of functionality that is just there and nice to use. And it also compiles down to the JVM, so you can mix java and kotlin classes and slowly migrate or at least try it out to see for yourself. At my work (banking environment) we are using it for Android and Backend and it has proven itself for the last years. Somewhere in the next years we might get java version >16, which will conclude same functionalities as Kotlin does. Also interesting for this topic -> https://youtu.be/te3OU9fxC8U

vcolin7: Thanks for the info Timo :) While I personally love the idea of working with Kotlin, using it in the Azure SDK for Android does not align with our current roadmap. We did spend quite a good while debating between using Java and Kotlin and are aware of the pros/cons, but ultimately decided to go with Java since it was better suited for our current needs. This does not mean that we are completely discarding the idea of supporting Kotlin in the SDK, it just might be a while before we reach that point.

PR #898, opened 2021-10-05
Updated Azure Core Logging's dependency on SLF4J from `slf4j-api` to `slf4j-android`.

vcolin7:
`slf4j-android` _["is basically a (i) repackaging of the SLF4J API part, together with (ii) a very lightweight binding implementation that simply forwards all SLF4J log requests to the logger provided on the Google Android platform."](http://www.slf4j.org/android/)_ --SLF4J Android, Logger name mapping.

This package includes an important workaround for the `Log.isLoggable()` Android API, [which will throw](https://stackoverflow.com/a/28168739) if a given tag is longer than 23 characters in length for Android API levels <= 25 [[1]](https://issuetracker.google.com/issues/124593220) [[2]](https://github.com/mvysny/slf4j-handroid/issues/2).

### DIFF
--- a/sdk/core/azure-core-logging/build.gradle
+++ b/sdk/core/azure-core-logging/build.gradle
@@ -11,7 +11,7 @@ android {
 
 dependencies {
     // <!-- begin: implementation Dependencies -->
-    implementation "org.slf4j:slf4j-api:$slf4jApiVersion"
+    implementation "org.slf4j:slf4j-android:$slf4jApiVersion"
     // <!-- end: implementation Dependencies -->
 
     // <!-- begin: test Dependencies -->


sdk/core/azure-core-logging/src/main/java/com/azure/android/core/logging/implementation/DefaultLogger.java
```diff
@@ -280,4 +283,73 @@ private String format(final String format, final Object arg1, final Object arg2)
...
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O || length <= TAG_MAX_LENGTH) {
+            return loggerName;
+        }
+
+        int tagLength = 0;
+        int lastTokenIndex = 0;
+        int lastPeriodIndex;
+        StringBuilder tagName = new StringBuilder(TAG_MAX_LENGTH + 3);
+
```
anuchandy, line 310: I guess the purpose of this method is to produce exactly the same shortened tag name as slf4jandroid? the slf4jandroid takes care of some additional edge cases - such as how to treat multiple dots, empty parts, etc. Just make sure we're not running into indexOutofBound in such cases.

In fact, in the past, I wrote a similar function for my own use case, just posting it here if it helps.

anuchandy:
```java

static final int TAG_MAX_LENGTH = 23;

public static String shortenTag(String tagName) {
    tagName = tagName.trim();
    if (tagName.length() <= TAG_MAX_LENGTH) {
        return tagName;
    }

    String [] parts = tagName.split("\\.");
    if (parts.length == 1) {
        return truncate(parts[0]);
    }

    StringBuilder shortenTagNameSb = new StringBuilder(23);
    for (int i = 0; i < parts.length - 1; i++) {
        final String part = parts[i].trim();
        if (part.length() == 0) {
            continue;
        }
        if (part.length() == 1) {
            shortenTagNameSb.append(part).append(".");
        } else {
            shortenTagNameSb.append(part.charAt(0)).append("*").append(".");
        }

        if (shortenTagNameSb.length() >= TAG_MAX_LENGTH) {
            return truncate(parts[parts.length - 1]);
        }
    }
    shortenTagNameSb.append(parts[parts.length - 1]);
    String shortenedTagName = shortenTagNameSb.toString();
    return shortenedTagName.length() >= TAG_MAX_LENGTH ? truncate(shortenedTagName) : shortenedTagName;
}

private static String truncate(String tagName) {
    return tagName.substring(0, TAG_MAX_LENGTH - 1) + "*";
}
```

anuchandy: Also, it would be great to have a couple of unit tests for the `loggerNameToTag` method:

Like: (Copied from slf4jAndroid main page)

```java
loggerNameToTag("org.example.project.MyClass")
loggerNameToTag("org.example.project.subproject.MyClass")
loggerNameToTag("org.example.MyQuiteLongNamedClassOfTooMuchCharacters")
loggerNameToTag("o.e.project.subproject.MyClass"));
loggerNameToTag("MyQuiteLongNamedClassNotInAPackage")
loggerNameToTag("o....e......project....subproject.MyClass")
```

sdk/core/azure-core-logging/build.gradle
```diff
@@ -11,7 +11,7 @@ android {
 
 dependencies {
     // <!-- begin: implementation Dependencies -->
-    implementation "org.slf4j:slf4j-api:$slf4jApiVersion"
+    implementation "org.slf4j:slf4j-android:$slf4jApiVersion"
```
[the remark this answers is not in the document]
vcolin7: Sounds good, thank you for the insight.

sdk/core/azure-core-logging/src/main/java/com/azure/android/core/logging/implementation/DefaultLogger.java
vcolin7: I actually used [SLF4J Android's implementation](https://github.com/twwwt/slf4j/blob/1e706fa4544bba7845a8964258015b85db35931c/slf4j-android/src/main/java/org/slf4j/impl/AndroidLoggerFactory.java#L44) for this tag sanitizer, so we are guaranteed to produce the same results as if a user were to use that library instead. The only difference is that I added a check for version number where anything above Android API level > 25 won't alter the tag, since the character limit was removed on Android Oreo (Android 26). Should I remove said limit?

vcolin7:
Added test cases and replaced the `slf4j-android` dependency with `slf4j-api` as suggested by @anuchandy.

PR #923, opened 2021-11-04
Enable encryption for push notification

CHILIU-MSFT:


sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/NotificationUtils.java
```diff
@@ -252,6 +266,60 @@ public static CloudType getUserCloudTypeFromSkypeToken(String skypeToken) {
         return CloudType.Public;
     }
 
+    public static boolean verifyEncryptedPayload(
+        byte[] f,
+        byte[] iv,
+        byte[] cipherText,
+        byte[] hmac,
+        SecretKey authKey) throws Throwable {
```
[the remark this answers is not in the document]
vcolin7: After doing some further reading, it looks like it's the encryption key, is this correct? If so, we should probably rename this parameter to something a bit more descriptive.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrarClient.java
```diff
@@ -39,8 +42,8 @@
     private static final String SKYPE_TOKEN_HEADER = "X-Skypetoken";
     private static final String USER_AGENT_HEADER = "User-Agent";
     private static final String CONTENT_TYPE_HEADER = "Content-Type";
-
     private static final String CONTENT_TYPE_JSON = "application/json";
+    private static final String CRYPTO_METHOD = "0x70";
```
vcolin7, line 46: What does this value stand for?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -207,9 +219,18 @@ private ChatEventType parsePushNotificationEventType(ChatPushNotification pushNo
     private ChatEvent parsePushNotificationEvent(ChatEventType chatEventType, ChatPushNotification pushNotification) {
         this.logger.info(" Try Jsonlize input.");
         JSONObject obj = new JSONObject(pushNotification.getPayload());
-        String jsonString = obj.toString();
-        this.logger.info(" Result: " + jsonString);
-        return NotificationUtils.toEventPayload(chatEventType, jsonString);
+        String decrypted;
+        try {
+            String encrypted = obj.getString("e");
+            this.logger.info(" Encrypted payload: " + encrypted);
```
vcolin7, line 225: Maybe we should log these contents at the debug level instead.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -207,9 +219,18 @@ private ChatEventType parsePushNotificationEventType(ChatPushNotification pushNo
...
+
+            decrypted = decryptPayload(encrypted);
+            this.logger.info(" Decrypted payload: " + decrypted);
```
vcolin7, line 228: Maybe we should log these contents at the debug level instead.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -284,4 +310,42 @@ public void run() {
...
+
+        this.previousCryptoKey = this.cryptoKey;
+        this.previousAuthKey = this.authKey;
+
+        this.cryptoKey = this.keyGenerator.generateKey();
+        this.authKey = this.keyGenerator.generateKey();
+    }
+
+    private String decryptPayload(String encryptedPayload) throws Throwable {
+        this.logger.info(" Decrypting payload.");
```
vcolin7, line 328: Is there a particular reason we add a space at the beginning of these logs?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -284,4 +310,42 @@ public void run() {
...
+
+        byte[] encryptedBytes = Base64Util.decodeString(encryptedPayload);
+        byte[] f = NotificationUtils.extractEncryptionKey(encryptedBytes);
```
vcolin7, line 331: Can we rename `f` to something like `encryptionKey`? Or is `f` known terminology in crypto like `iv`?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -284,4 +310,42 @@ public void run() {
...
+        if (NotificationUtils.verifyEncryptedPayload(f, iv, cipherText, hmac, this.authKey)) {
+            return NotificationUtils.decryptPushNotificationPayload(iv, cipherText, this.cryptoKey);
+        }
+
+        // When client has registered a new key, because of eventual consistency, latencies and concurrency,
+        // the old key can still be used by server side for some notifications
+        // Try old key if the new key failed to decrypt the payload.
+        if (NotificationUtils.verifyEncryptedPayload(f, iv, cipherText, hmac, this.previousAuthKey)) {
+            return NotificationUtils.decryptPushNotificationPayload(iv, cipherText, this.previousCryptoKey);
+        }
```
vcolin7, line 345: Is it possible to use an old key that has expired a long time ago? How often are we expecting users to call the function to renew?

vcolin7 (commented):
The encryption logic looks good, I added a few comments on other areas though :)

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
CHILIU-MSFT: The reason is if the class name is too long, in log this.logger.info() will be printed directly after class name without a space. So I manually added a space here to separate the class name & log messages...

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
CHILIU-MSFT: For renew, I am following this Recommended rotation process https://skype.visualstudio.com/DefaultCollection/SCC/_wiki/wikis/SCC.wiki/2253/Cryptography?anchor=recommended-rotation-process

Currently, key rotation happens when registration is renewed, which is 90000 seconds (25 hours). In the doc, they also mentioned I should "Start a short countdown for a grace period". I'll update to this.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -322,30 +324,44 @@ private void refreshEncryptionKeys() throws Throwable {
...
         // Try old key if the new key failed to decrypt the payload.
-        if (NotificationUtils.verifyEncryptedPayload(f, iv, cipherText, hmac, this.previousAuthKey)) {
+        if (inKeyRotationGracePeriod()
+            && NotificationUtils.verifyEncryptedPayload(encryptionKey, iv, cipherText, hmac, this.previousAuthKey)) {
             return NotificationUtils.decryptPushNotificationPayload(iv, cipherText, this.previousCryptoKey);
         }
 
         // Reject the push when computed signature does not match the included signature - it can not be trusted
         throw logger.logExceptionAsError(
             new RuntimeException("Invalid encrypted push notification payload! Drop the request."));
```
vcolin7, line 354: I think we should throw an exception with a different error message if the problem was due to an invalid key, as we really don't know if the payload was valid or not and saying so is misleading for users and makes things hard to debug.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -351,7 +351,7 @@ private String decryptPayload(String encryptedPayload) throws Throwable {
 
         // Reject the push when computed signature does not match the included signature - it can not be trusted
         throw logger.logExceptionAsError(
-            new RuntimeException("Invalid encrypted push notification payload! Drop the request."));
+            new RuntimeException("Invalid or expired authorization key. Drop the request!"));
```
vcolin7, line 354: What I meant to suggest with my previous comment was to provide an accurate error message for whatever case we encounter, in other words, we need to tell the user why the payload could not be verified, either because the key is expired or because the payload could not be verified with a valid key. So we need at least two error messages in my opinion. What do you think?

CHILIU-MSFT: Ah. Got it. Here what we are doing is to calculate a signature of the encrypted payload using local stored authKey, and compare the local calculated signature with the one the client received from the server. If does not match, we would think the encrypted payload can not be trusted.

Here, we would always think the local key in client side is valid, and assume the encrypted payload broken or attacked by someone. It should always be "the payload could not be verified with a valid key".

Do you have any suggestion whether we have better error message here? Thanks!

vcolin7: I agree that if the payload cannot be verified we should assume the key is valid and the problem relies with the payload. I was thinking that about the scenario where the grace time for renewing the key expires, it doesn't feel right saying that the payload was invalid in such a case, as developers wouldn't know they need to renew the key to attempt the verify operation. So I would either have one distinct error message for each of those two cases, or phrase this as the following:

```suggestion
            new RuntimeException("Invalid encrypted push notification payload or expired authorization key. Dropped the request!"));
```

Also, I'm not sure about using exclamation signs (!) at the end of these messages but honestly that's not a big deal for the time being, in my opinion.

CHILIU-MSFT: For key renewing scenario, it would always be the SDK itself to create a new key for registrar request, then mark the old keys as previousKeys and start the grace time that still try the previousKeys if new key failed in validation.

CHILIU-MSFT: We don't have the case for "expired authorization key" here.

vcolin7: Maybe the key itself doesn't expire after the "grace" period passes, but would we allow customers to use a very old key? If not, we need to be clear in letting them know they need to provide a more recent one. If the key is not all that old and the payload cannot be verified, then we let customers know the payload is not trustworthy.

CHILIU-MSFT: Here keys are auto-generated & rotated by SDK (not provided by customer) every 25 hours together with registration renew. Would this be an issue that we should let customer pass in their own keys? If this is OK, we don't have "very old key" scenario. Unable to verify means the payload is not trustworthy.

vcolin7: Oh I see, for some reason I thought these keys could come from users. If the SDK is taking care of all this, then probably just telling users the payload could not be verified is enough.

vcolin7 (approved):

PR #1016, opened 2022-02-16
Updated Jackson version to 2.12.6 in all modules.

vcolin7:
Cannot upgrade to `2.13+` because of #1017, so we are upgrading to the latest version below.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -264,7 +264,7 @@ allprojects {
         appCompatVersion = "1.2.0"
         azureCommunicationCommonVersion = "1.0.1"
         azureCoreVersion = "1.0.0-beta.9"
-        jacksonVersion = "2.12.5" // Do not upgrade to 2.13, as it introduced using an API not available in javax.xml.stream:stax-api:1.0-2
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -264,7 +264,7 @@ allprojects {
         appCompatVersion = "1.2.0"
         azureCommunicationCommonVersion = "1.0.1"
         azureCoreVersion = "1.0.0-beta.9"
-        jacksonVersion = "2.12.5" // Do not upgrade to 2.13, as it introduced using an API not available in javax.xml.stream:stax-api:1.0-2
+        jacksonVersion = "2.12.6" // Do not upgrade to 2.13, as it introduced using an API not available in javax.xml.stream:stax-api:1.0-2
```
joshfree, line 267: is there a bug tracking the 2.13 issue that can be added to this comment?

joshfree (approved):

build.gradle
vcolin7: Yes, I created #1017 and will add it to the comment in our `build.gradle` file.

PR #1078, opened 2022-04-11
Bumping version post release Apr

jiminwen-msft:
Bumping version post release Apr


CHILIU-MSFT:
We can just leave the empty google-services.json file there.

jiminwen-msft:
> We can just leave the empty google-services.json file there.

Updated

CHILIU-MSFT (approved):

PR #1097, opened 2022-05-02
Jimin/dev2

jiminwen-msft:
task: https://skype.visualstudio.com/SPOOL/_workitems/edit/2823275

This PR tries to introduce a reliable approach of refreshing push notification token using worker manager. There changes to both SDK and sample app. Contoso APP is required to make configuration change in order to run WorkManager in background. Otherwise, we fall back to use the TimerTask in order not to breaking the APP.

For SDK:
1. RenewalTokenWorker: The worker that does the job in background. 
2. RenewalTokenWorkerFactory: Factory for the worker. It is created because we could not rely on the default factory. We need customized factory which injects the CommunicationCredential. And this CommunicationCredential needs to be injected from Tontoso APP during the initiation phase of the APP.
3. RenewalWorkerDataContainer: Data container for updating and storing secrete keys shared by PushNotificationClient and RenewalTokenWorker. Output Data from periodic request is empty by design. Storing the key in memory or in database are only solutions to pass data/result. This class is singleton class so that it remains in the memory. Even if APP closed and memory gets cleared. This class has KeyGenerator which could still handle the work in this situation.
4. Gradle: inject worker related dependency 

For Contoso. Configuration changes are required to use the WorkerManager. Otherwise, SDK would fall back to use the TimedTask to do the action.
1. Create customized configuration to use RenewalTokenWorkerFactory
2. manifest change to refer to the configuration
3. Set the enable worker manager flag 
4.  Gradle: inject worker related dependency


sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RenewalWorkerDataContainer.java
```diff
@@ -0,0 +1,78 @@
...
+    @Getter
+    @Setter
+    private long keyRotationTime;
+
+    public RenewalWorkerDataContainer()  {
+        try {
+            this.keyGenerator = KeyGenerator.getInstance("AES");
+        } catch (NoSuchAlgorithmException e) {
+            e.printStackTrace();
+            throw new RuntimeException("KeyGenerator failed: " + e.getMessage());
```
vcolin7, line 53: Wouldn't throwing here make the customer application crash?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RenewalWorkerDataContainer.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Getter
+    @Setter
+    private long keyRotationTime;
+
+    public RenewalWorkerDataContainer()  {
+        try {
+            this.keyGenerator = KeyGenerator.getInstance("AES");
+        } catch (NoSuchAlgorithmException e) {
+            e.printStackTrace();
```
vcolin7, line 52: We should use [`ClientLogger.logExceptionAsError()`](https://github.com/Azure/azure-sdk-for-android/blob/9fc5dbc53de69a20d9009a5bad9cd9ff0392ff54/sdk/core/azure-core-logging/src/main/java/com/azure/android/core/logging/ClientLogger.java#L206) instead.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RenewalWorkerDataContainer.java
```diff
@@ -0,0 +1,78 @@
...
+    public static RenewalWorkerDataContainer instance() {
+        if (renewalWorkerDataContainer == null) {
+            renewalWorkerDataContainer = new RenewalWorkerDataContainer();
+        }
+        return renewalWorkerDataContainer;
+    }
+
+    /**
+     * Update stored cryptoKeys and authKeys. The current secret key become the new values generated from key generator.
+     * The previous cryptoKeys and authKeys store the previous values. Those keys are used for decrypt push notification
```
vcolin7, line 68: I think the comment is a bit confusing: by reading this I could think the previous keys are used for decryption instead of the current ones.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RenewTokenWorker.java
```diff
@@ -0,0 +1,78 @@
...
+
+import com.azure.android.communication.chat.implementation.notifications.NotificationUtils;
+import com.azure.android.communication.common.CommunicationTokenCredential;
+
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Worker for renewing registra token. This worker could execute when APP closed. This is introduced to allow renewal
+ * happen even when APP gets closed.
+ * https://skype.visualstudio.com/SPOOL/_queries/query/af989357-74be-49b6-833a-d4359cd1cbd4/
```
vcolin7, line 21: Do we need this link in the documentation?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RenewTokenWorker.java
```diff
@@ -0,0 +1,78 @@
...
+        this.registrarClient = new RegistrarClient();
+        this.communicationTokenCredential = communicationTokenCredential;
+        this.renewalWorkerDataContainer = RenewalWorkerDataContainer.instance();
+    }
+
+    @NonNull
+    @Override
+    public Result doWork() {
+        int attempts = getRunAttemptCount();
+        Log.i("RenewTokenWorker", "RenewTokenWorker execution in background: " + attempts);
```
vcolin7, line 46: We should use `ClientLogger` to log accross our libraries, it uses the Android logger by default.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -313,40 +375,25 @@ public void run() {
...
-        this.cryptoKey = this.keyGenerator.generateKey();
-        this.authKey = this.keyGenerator.generateKey();
-
-        this.keyRotateTimeMillis = System.currentTimeMillis();
-    }
-
     private String decryptPayload(String encryptedPayload) throws Throwable {
         this.logger.verbose(" Decrypting payload.");
-
+        Log.i("decryption", "cypto: " + renewalWorkerDataContainer.getCurAuthKey() + ", auth: " + renewalWorkerDataContainer.getCurAuthKey());
```
vcolin7, line 380: We prefer to use `ClientLogger` as it already uses the Android logger by default unless otherwise specified.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -235,7 +262,42 @@ private ChatEvent parsePushNotificationEvent(ChatEventType chatEventType, ChatPu
...
+                if (failed) {
+                    RuntimeException exception = new RuntimeException(
+                        "Registration renew request failed after "
+                            + NotificationUtils.MAX_REGISTRATION_RETRY_COUNT
+                            + "retries.");
+                    this.registrationErrorHandler.accept(exception);
+                    stopPushNotifications();
+                    this.logger.info("Renew token failed");
+                } else {
+                    this.logger.info("Renew token succeeded");
```
vcolin7, line 294: If we use the class `ClientLogger` here it could mean that when an app is closed and re-opened we'd potentially have a second logger until the worker finally stops, right @anuchandy? I guess it wouldn't be a big deal as long as we don't create more than one worker for this type of work, but I'd still like to hear your opinion on this.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -395,4 +395,11 @@ public void addEventHandler(ChatEventType chatEventType, RealTimeNotificationCal
     public void removeEventHandler(ChatEventType chatEventType, RealTimeNotificationCallback listener) {
         this.signalingClient.off(chatEventType, listener);
     }
+
+    /**
+     * Enable worker manager which renews registra token
+     */
+    public void enableWorkerManager() {
+        this.pushNotificationClient.setEnableWorkManager(true);
+    }
```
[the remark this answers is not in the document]
JonathanGiles: We support both approaches - with boolean and without - it just depends on use case and readability. With only what I see in front of me, my bigger question is why are we allowing state changes in the client? We typically strive to avoid this and have all state set in the builder.

PR #1123, opened 2022-05-26
Jimin/dev5 key

jiminwen-msft:
task: 
1. https://skype.visualstudio.com/SPOOL/_workitems/edit/2823275
2. https://skype.visualstudio.com/SPOOL/_workitems/edit/2867798

Related PR: https://github.com/Azure/azure-sdk-for-android/pull/1097

This PR tries to 
1. Introduce a reliable approach of refreshing push notification registration using worker manager. There are changes to both SDK and sample app. Contoso APP is required to make configuration change in order to run WorkManager in background. Otherwise, client faces RuntimeException. 
2. Introduce key-store to persist secrete keys. Secrete keys are used for push notification registration and decrypt push notification payloads. Using key-store ensure the keys are persisted even APP gets closed. And instead of storing two pairs of secrete keys. We change to store 10 pairs of keys. And try all of them when decrypt push notification payload.
4. We set registration renewal work to repeat every 15 minutes. And set TTL of push notification registration as 45 minutes. So that we have 3 chances of execute worker before it expires.
_________
More specific changes: 
- SDK side:
1. RegistrationRenewalWorker: The worker that renews registration.
2. RegistrationRenewalWorkerFactory : Factory for the worker. It is created because we could not rely on the default factory. We need customized factory which injects the CommunicationCredential. And this CommunicationCredential needs to be injected from Tontoso APP during the initiation phase of the APP.
3. RegistrationDataContainer : Data container for updating and storing secrete keys shared by PushNotificationClient and RenewalTokenWorker. It wraps key-store inside. So that it could persist all pairs of secrete keys in file system.
Gradle: inject worker related dependency

For Contoso. Configuration changes are required to use the WorkerManager. 
1. Create customized configuration to use RenewalTokenWorkerFactory
2. manifest change to refer to the configuration
3. Gradle: inject worker related dependency


angellan-msft (approved):

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -77,34 +78,20 @@ public boolean hasStarted() {
...
-        try {
-            this.refreshEncryptionKeys();
-            this.registrarClient.register(skypeUserToken, deviceRegistrationToken, this.cryptoKey, this.authKey);
-        } catch (Throwable throwable) {
-            this.logger.error("Failed to start push notifications!");
-            errorHandler.accept(throwable);
-            return;
-        }
+        this.workManager = WorkManager.getInstance();
+        this.registrationDataContainer = RegistrationDataContainer.instance();
```
xerxestph, line 86: Does registrationDataContainer need to be singleton? If so, why do we need to keep a reference of it?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -77,34 +78,20 @@ public boolean hasStarted() {
...
 
         this.isPushNotificationsStarted = true;
         this.pushNotificationListeners.clear();
         this.logger.info("Successfully started push notifications!");
-
-        long delayInMS = 1000L * (long) (Integer.parseInt(PUSHNOTIFICATION_REGISTRAR_SERVICE_TTL) - 30);
-        this.scheduleRegistrationRenew(delayInMS, 0);
+        // 15 minutes is the minimum interval worker manager allows
+        long delay = 15;
+        this.startRegistrationRenewalWorker(delay);
```
xerxestph, line 94: I assume that a registration request will be sent immediately and then it will wait for 15mins before the next renewal, right? in this case calling it "delay" could be confusing. "interval" might be a better name.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -235,133 +225,59 @@ private ChatEvent parsePushNotificationEvent(ChatEventType chatEventType, ChatPu
...
+        this.logger.info("Initialize RegistrationRenewalWorker in background");
+
+        Data inputData = new Data.Builder().putString("deviceRegistrationToken", deviceRegistrationToken).build();
+
+        PeriodicWorkRequest renewTokenRequest =
+            new PeriodicWorkRequest.Builder(RegistrationRenewalWorker.class, intervalInMinutes, TimeUnit.MINUTES)
+                .setBackoffCriteria(BackoffPolicy.LINEAR, 10, TimeUnit.SECONDS)
+                .setInputData(inputData)
+                .build();
+        workManager.enqueueUniquePeriodicWork("Renewal registration", ExistingPeriodicWorkPolicy.REPLACE, renewTokenRequest);
```
xerxestph, line 238: Perhaps a more specific name to avoid name collision?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationRenewalWorker.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        // Registration
+        Data inputData = getInputData();
+        String skypeUserToken;
+        String deviceRegistrationToken =  inputData.getString("deviceRegistrationToken");
+        try {
+            skypeUserToken = communicationTokenCredential.getToken().get().getToken();
+            refreshCredentials();
+            Pair<SecretKey, SecretKey> keyPair = registrationDataContainer.getLastPair();
```
xerxestph, line 67: There was a feedback for iOS on the use of the term "keyPair" here - usually when we talk about key pair, we think about asymmetric keys. We should find a better name for this.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
```diff
@@ -235,133 +225,59 @@ private ChatEvent parsePushNotificationEvent(ChatEventType chatEventType, ChatPu
...
+
+        //Checking the result of last execution. There are two states for periodic work: ENQUEUED and RUNNING. We do checking
+        //for every ENQUEUED state, meaning last execution has completed.
+        workManager.getWorkInfoByIdLiveData(renewTokenRequest.getId()).observeForever(workInfo -> {
```
xerxestph, line 242: Does this callback get invoked only if the app is running?

xerxestph: How does Contoso app get notified when push notification renewal has failed?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
jiminwen-msft: Yes. It is interval. The first execution starts immediately.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
jiminwen-msft: It is only invoked when APP is alive. Currently the client is not able to know there is no active registration.

jiminwen-msft: What comes to my mind is introducing foreground notification. I think that is the only way to notify clients. But is seems to be over-kill of the problem and I have not analyzed if it is guaranteed to execute in background.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
jiminwen-msft: Implementing with a singleton is much safter and easier. The credential keys are loaded from file system and gets refreshed every 15 minutes. The refresh method is invoked by the worker every 15 minutes. So I am using singleton to ensure the same key-store is being shared/used. 

Do you prefer using RegistrationDataContainer.instance() instead of keeping a reference? I just feel having a reference avoid duplicate calls to this method RegistrationDataContainer.instance().

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
xerxestph: Is it possible for Contoso to provide a callback method / class for our Worker class to call directly when renewal has failed?

jiminwen-msft: As what has been discussed offline, Contoso will pass the exception hander to the registration worker.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java
```diff
@@ -0,0 +1,125 @@
...
+    public RegistrationKeyStore() {
+        map = new HashMap<>();
+    }
+
+    public int getSize() {
+        return map.size();
+    }
+
+    //Loading data with given path to memory
+    public void loading(InputStream inputStream) {
```
xerxestph, line 53: please rename to "load"

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java
```diff
@@ -0,0 +1,125 @@
...
+            Type type = new TypeToken<Map<String, RegistrationKeyEntry>>() {}.getType();
+            Reader reader = new InputStreamReader(inputStream);
+            map = gson.fromJson(reader, type);
+            reader.close();
+        } catch (IOException e) {
+            clientLogger.logExceptionAsError(new RuntimeException(e));
+        }
+    }
+
+    public void storingKeyEntry(String alias, String path, RegistrationKeyEntry registrationKeyEntry) {
```
xerxestph, line 69: please rename to "storeKeyEntry"

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java
```diff
@@ -0,0 +1,125 @@
...
+ *         "creationTime": "1234";
+ *     },
+ *     "alias_2": {
+ *         "credential": "abcdefgh",
+ *         "creationTime": "1234";
+ *     }
+ * }
+ */
+public class RegistrationKeyStore {
+    private Map<String, RegistrationKeyEntry> map;
```
xerxestph, line 40: Not sure why an alias is needed. Can this be just a list of keys?

jiminwen-msft: The major reason is that we need to distinguish cryptoKeys and authKeys. Having a key with different prefix is an easy way to achieve this. Using list structure probably requires managing separate files for each of them.

xerxestph: aren't the crypto and auth keys always go in pairs and share the same expiry time? if so, why don't we have a structure like this:
```json
[
  { "crypt": "...", "auth": "...", expiryTime: ... },
  { "crypt": "...", "auth": "...", expiryTime: ... },
  { "crypt": "...", "auth": "...", expiryTime: ... }
]
```

jiminwen-msft: Let me trying implementing this

jiminwen-msft: Changed to list

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,144 @@
...
+    private void load(String path) {
+        try (FileInputStream fis = new File(path).exists() ? new FileInputStream(path) : null) {
+            registrationKeyStore.load(fis);
+            Log.v("RegistrationContainer", "key-store size: " + registrationKeyStore.getSize());
+        } catch (Exception e) {
+            clientLogger.logExceptionAsError(new RuntimeException("Failed to load key store", e));
+        }
+    }
+
+    private String secreteKeyToStr(SecretKey secretKey) {
```
xerxestph, line 126: typo: secrete vs. secret

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java
```diff
@@ -0,0 +1,154 @@
...
+        stack.addAll(list);
+        return stack;
+    }
+
+    public static class RegistrationKeyEntry {
+        private String cryptoCredential;
+
+        private String authCredential;
+
+        private long creationTime;
```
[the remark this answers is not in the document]
jiminwen-msft: I think tracking creation time is easier here. 1. Tracking expiration time requires us to update every entry during refresh. 2. If we change TTL, we just need to change the value of EXPIRATION_TIME_MINUTES  in RegistrationKeyManager.

angiurgiu (commented):
Seems like you have some checkstyle violations:

[ant:checkstyle] [ERROR] /mnt/vss/_work/1/s/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java:7:8: Unused import - android.util.Pair. [UnusedImports]
[ant:checkstyle] [ERROR] /mnt/vss/_work/1/s/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java:35:8: Unused import - javax.crypto.spec.SecretKeySpec. [UnusedImports]
[ant:checkstyle] [ERROR] /mnt/vss/_work/1/s/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java:19:8: Unused import - java.util.Deque. [UnusedImports]
[ant:checkstyle] [ERROR] /mnt/vss/_work/1/s/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java:20:8: Unused import - java.util.LinkedList. [UnusedImports]
[ant:checkstyle] [ERROR] /mnt/vss/_work/1/s/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java:67:69: '{' is not followed by whitespace. [WhitespaceAround]
[ant:checkstyle] [ERROR] /mnt/vss/_work/1/s/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java:67:70: '}' is not preceded with whitespace. [WhitespaceAround]
[ant:checkstyle] [ERROR] /mnt/vss/_work/1/s/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java:87:9: 'try' is not followed by whitespace. [WhitespaceAround]
[ant:checkstyle] [ERROR] /mnt/vss/_work/1/s/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationRenewalWorker.java:8:8: Unused import - android.util.Pair. [UnusedImports]
[ant:checkstyle] [ERROR] /mnt/vss/_work/1/s/sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationRenewalWorker.java:22:8: Unused import - javax.crypto.SecretKey. [UnusedImports]

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java
```diff
@@ -0,0 +1,154 @@
...
+            bos.close();
+            fos.close();
+        } catch (IOException e) {
+            clientLogger.logExceptionAsError(new RuntimeException("Filed writing key map to file", e));
+        }
+    }
+
+    public void storeKeyEntry(String path, RegistrationKeyEntry registrationKeyEntry) {
+        list.add(list.size(), registrationKeyEntry);
+        writeJsonToFile(path);
```
vcolin7, line 102: How often do we store a new key entry?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    public void deleteFirstEntry(String path) {
```
vcolin7, line 105: Why would we want to delete the first entry?

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,144 @@
...
+import javax.crypto.SecretKey;
+
+/**
+ * A singleton wrapper which manage the persistence and rotation of secrete keys and execution result related to #{RenewalTokenWorker}.
+ * The secrete keys are persisted and loaded using key-store API. We generate new secrete keys every refresh and evict expired keys
+ * from key-store every refresh.
+ *
+ * RegistrationKeyStore is used as object to manage the keys in file/memory.
+ */
+public final class RegistrationKeyManager {
```
vcolin7, line 25: As a general note, we should have run approach by the Crypto Board. The folks working on this feature for iOS were told the same and if I recall correctly, they were going to schedule a meeting with them. At first glance what you're doing seems right to me but it's better to check with the Board to ensure we are not putting customer data at risk.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,144 @@
...
+    //The duration we persist keys in key-store
+    public static final int EXPIRATION_TIME_MINUTES = 45;
+
+    private ClientLogger clientLogger = new ClientLogger(RegistrationKeyManager.class);
+
+    private RegistrationKeyManager()  {
+        try {
+            this.keyGenerator = KeyGenerator.getInstance("AES");
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException("KeyGenerator failed: " + e.getMessage());
```
vcolin7, line 41: Don't throw exceptions directly, log them using `logExceptionAsError()` in `ClientLogger` and then throw the Exception returned by said method.
```suggestion
            throw clientLogger.logExceptionAsError(new RuntimeException("KeyGenerator failed: " + e.getMessage()));
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    private int getNumOfPairs() {
+        int numKeyPairs = 0;
+        try {
+            numKeyPairs = registrationKeyStore.getSize() / 2;
+        } catch (Exception e) {
+            clientLogger.logExceptionAsError(new RuntimeException("Failed to get size from key store", e));
+        }
+        Log.v("RegistrationContainer", "Number of pairs in key-store: " + numKeyPairs);
```
vcolin7, line 69: Log using `ClientLogger`.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,144 @@
...
+            RegistrationKeyStore.RegistrationKeyEntry entry = new RegistrationKeyStore.RegistrationKeyEntry(cryptoCredential, authCredential, currentTimeMillis);
+            registrationKeyStore.storeKeyEntry(path, entry);
+        }
+    }
+
+    private void rotateKeysBasedOnTime(String path) {
+        int removedEntry = 0;
+        int size = getNumOfPairs();
+        //Delete expired keys
+        for (int curIndex = 1; curIndex <= size; curIndex++) {
```
vcolin7, line 95: Why are we starting with 1 and using `<=`, as opposed to starting with 0 and using `<`? I think the error message below where `curIndex` is used could be confusing otherwise.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,144 @@
...
+                } catch (Exception e) {
+                    clientLogger.logExceptionAsError(new RuntimeException("Failed to delete entry from key-store with index: " + curIndex));
+                }
+                removedEntry++;
+            } else {
+                break;
+            }
+        }
+
+        Log.v("RegistrationKeyManager", "Removed " + removedEntry + " of entries");
```
vcolin7, line 113: Log using `ClientLogger`.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    //Loads data from key store file to key store entry
+    private void load(String path) {
+        try (FileInputStream fis = new File(path).exists() ? new FileInputStream(path) : null) {
+            registrationKeyStore.load(fis);
+            Log.v("RegistrationContainer", "key-store size: " + registrationKeyStore.getSize());
```
vcolin7, line 120: Log using `ClientLogger`.

samples/sample-chat-app/build.gradle
```diff
@@ -46,4 +46,7 @@ dependencies {
     implementation platform('com.google.firebase:firebase-bom:28.1.0')
     implementation 'com.google.firebase:firebase-analytics'
     implementation 'com.google.firebase:firebase-messaging'
+    def work_version = "2.7.1"
+    // (Java only)
+    implementation "androidx.work:work-runtime:$work_version"
```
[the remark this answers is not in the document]
jiminwen-msft: Users need to 1. Add this dependency 2. Disable default worker factory initializer 3. Add MyAppConfiguration which is the costumed worker factory initializer

sdk/communication/azure-communication-chat/build.gradle
```diff
@@ -58,4 +58,10 @@ dependencies {
 
     testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitJupiterVersion"
     // <!-- end: test Dependencies -->
+    def work_version = "2.7.1"
+    // (Java only)
+    implementation "androidx.work:work-runtime:$work_version"
```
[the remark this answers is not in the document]
jiminwen-msft: I think this is still needed on sample code side. As the MyAppConfiguration needs to import WorkerManager. There is compiling exception without adding it on sample code side.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyStore.java
jiminwen-msft: We renew the registration every 15 minutes. And will add 1-3 entries based on retry attempts. So the TTL is set as 45 minutes (2700 seconds). Registra team is going to change the TTL on their to around 50 minutes.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
jiminwen-msft: I am not sure of the reason. But log using client logger does not log the message in the "run tab". I am adding log here mainly to help debugging.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrarClient.java
```diff
@@ -237,7 +236,7 @@ private void addRequestBody(
         transport.creationTime = "";
         transport.context = "";
         transport.path = deviceRegistrationToken;
-        transport.ttl = PUSHNOTIFICATION_REGISTRAR_SERVICE_TTL;
+        transport.ttl = "2700";
```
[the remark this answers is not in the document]
jiminwen-msft: Registra team is going to reduce the TTL on their side to around 50 minutes. So we are reducing the TTL value on our side to 45 minutes.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
```diff
@@ -315,15 +315,14 @@ public void stopRealtimeNotifications() {
      * If there's an error during registration initialization or renewal,
      * push notifications will be disabled and {@code errorHandler} will be called.
      * @param deviceRegistrationToken Device registration token obtained from the FCM SDK.
-     * @param errorHandler error handler callback for registration failures
      * @throws RuntimeException if push notifications failed to start.
      */
-    public void startPushNotifications(String deviceRegistrationToken, Consumer<Throwable> errorHandler) {
+    public void startPushNotifications(String deviceRegistrationToken) {
```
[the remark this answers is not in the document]
jiminwen-msft: We are passing this handler to make the exception being handled inside the worker. The logic in worker is guaranteed to execute. Otherwise, we are using listener in pushnotification client to handle exception. The listener is not ensured to execute when the thread is idle/killed.

sdk/communication/azure-communication-chat/build.gradle
```diff
@@ -58,4 +58,10 @@ dependencies {
...
+
+    implementation 'com.google.code.gson:gson:2.8.5'
```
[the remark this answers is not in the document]
jiminwen-msft: GSON is used in RegistrationKeyStore to convert between list of entries and JSON. And credentials are persisted in device local directory.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationRenewalWorkerFactory.java
```diff
@@ -0,0 +1,33 @@
...
+    private Consumer<Throwable> exceptionHandler;
+
+    public RegistrationRenewalWorkerFactory(CommunicationTokenCredential communicationTokenCredential, Consumer<Throwable> exceptionHandler) {
+        this.communicationTokenCredential = communicationTokenCredential;
+        this.exceptionHandler = exceptionHandler;
+    }
+
+    @Nullable
+    @Override
+    public ListenableWorker createWorker(@NonNull Context appContext, @NonNull String workerClassName, @NonNull WorkerParameters workerParameters) {
```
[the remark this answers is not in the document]
jiminwen-msft: The workerClassName is inherent from the interface. So I would keep it though is not being used for now. 

I used enqueueUniqueWork method, which keeps only one worker being active. The newly added worker would replace the old one.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/PushNotificationClient.java
jiminwen-msft: Plan to retain this logic to not breaking this API. Will create a new PR adding a new API without the exception handler.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
jiminwen-msft: I think this is safe approach. We are persisting each credential for 45 minutes. And those credentials are not notification data. They are the secreteKey to decrypt the actual data. The push notification data itself are never being persisted in the device.

sdk/communication/azure-communication-chat/build.gradle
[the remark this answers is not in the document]
vcolin7: I don't know the exact policy on bringing on external libraries. Maybe @JonathanGiles could chime in here. In any case, if we end up keeping it around we should add it to [this section](https://github.com/Azure/azure-sdk-for-android/blob/d551097cc3d216a9b152b9a7ce894b952b3da9d4/build.gradle#L262) of our root `build.gradle` file to have a centralized place to keep track of our external dependencies.

sdk/communication/azure-communication-chat/build.gradle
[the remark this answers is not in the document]
vcolin7: And I assume we also need this in the library `build.gradle`, right?

samples/sample-chat-app/build.gradle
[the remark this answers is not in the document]
vcolin7: But do they need to do this even if they don't intend on using the push notifications functionality? I would assume not.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatAsyncClient.java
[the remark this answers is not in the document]
vcolin7: Based on a separate discussion, we will be keeping this API around but will eventually deprecate it.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/ChatClient.java
```diff
@@ -223,11 +223,10 @@ public void stopRealtimeNotifications() {
      * If there's an error during registration initialization or renewal,
      * push notifications will be disabled and {@code errorHandler} will be called.
      * @param deviceRegistrationToken Device registration token obtained from the FCM SDK.
-     * @param errorHandler error handler callback for registration failures
      * @throws RuntimeException if push notifications failed to start.
      */
-    public void startPushNotifications(String deviceRegistrationToken, Consumer<Throwable> errorHandler) {
-        client.startPushNotifications(deviceRegistrationToken, errorHandler);
+    public void startPushNotifications(String deviceRegistrationToken) {
```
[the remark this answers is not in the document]
vcolin7: Based on a separate discussion, we will be keeping this API around but will eventually deprecate it.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrarClient.java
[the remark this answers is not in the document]
vcolin7: Is this the only place where this value will be used? If so, I think this should be fine then.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
vcolin7: We should still run this by the Crypto Board. You can send them an email directly or coordinate with the iOS folks if they already have a conversation going or set up a meeting.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,152 @@
...
+
+    private ClientLogger clientLogger = new ClientLogger(RegistrationKeyManager.class);
+
+    private RegistrationKeyManager()  {
+        try {
+            this.keyGenerator = KeyGenerator.getInstance("AES");
+        } catch (NoSuchAlgorithmException e) {
+            RuntimeException runtimeException = new RuntimeException("KeyGenerator failed: " + e.getMessage());
+            clientLogger.logExceptionAsError(runtimeException);
+            throw runtimeException;
```
vcolin7, line 45: The [guidelines](https://azure.github.io/azure-sdk/android_implementation.html#android-logging-log-and-throw) indicate we should throw the `RuntimeException` returned by `logExceptionAsError`.
```suggestion
            throw clientLogger.logExceptionAsError(new RuntimeException("KeyGenerator failed: " + e.getMessage()));
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
vcolin7: I'll look into why it doesn't show in the main console, but it definitely shows in LogCat.

vcolin7 (commented):
Added more comments :)

sdk/communication/azure-communication-chat/build.gradle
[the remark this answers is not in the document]
JonathanGiles: All external dependencies need my approval before their first use in our libraries.

sdk/communication/azure-communication-chat/build.gradle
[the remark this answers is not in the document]
jiminwen-msft: Yes. It is needed on both SDK end and sample APP end

sdk/communication/azure-communication-chat/build.gradle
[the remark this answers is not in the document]
jiminwen-msft: Hello. @JonathanGiles. Could you review this use case? GSON is being widely used for transform between JSON and objects. The risk using it seems low.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
jiminwen-msft: Seems the same as my implementation. I am doing it in two lines of codes. But I will change it as this.

sdk/communication/azure-communication-chat/build.gradle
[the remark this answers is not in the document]
jiminwen-msft: https://github.com/search?p=3&q=org%3AAzure+com.google.code.gson%3Agson&type=Code. Did a search just now. This library is used by many Azure team as well. For example azure-communication-ui.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -64,84 +94,266 @@ public boolean getLastExecutionSucceeded() {
...
+    }
+
+    public SecretKey getSecreteKey(String alias) {
+        KeyStore.ProtectionParameter protParam =
+            new KeyStore.PasswordProtection(getPassword());
+
+        // get my private key
+        KeyStore.SecretKeyEntry pkEntry = null;
+        try {
+            pkEntry = (KeyStore.SecretKeyEntry) keyStore.getEntry(alias, protParam);
```
vcolin7, line 273: I don't think this will play nice if the result of `getEntry()` is null. I think we should check for null before attempting casting to `KeyStore.SecretKeyEntry`.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -64,84 +94,266 @@ public boolean getLastExecutionSucceeded() {
...
+            keyCreationTimeStore.load(fis2);
         } catch (IOException e) {
-            clientLogger.logExceptionAsError(new RuntimeException("Failed to load key store", e));
+            throw clientLogger.logExceptionAsError(new RuntimeException("Failed to load key creation time store", e));
         }
     }
 
-    private String secretKeyToStr(SecretKey secretKey) {
-        return Base64Util.encodeToString(secretKey.getEncoded());
+    private char[] getPassword() {
```
vcolin7, line 225: What's the difference between this and `getKeyStorePassword()`?

vcolin7: Also, same question about using a hard-coded value.

vcolin7 (commented):
Added some comments in general and a couple about the use of `KeyStore`.

samples/sample-chat-app/build.gradle
```diff
@@ -46,4 +46,6 @@ dependencies {
     implementation platform('com.google.firebase:firebase-bom:28.1.0')
     implementation 'com.google.firebase:firebase-analytics'
     implementation 'com.google.firebase:firebase-messaging'
+    def work_version = "2.7.1"
+    implementation "androidx.work:work-runtime:$work_version"
```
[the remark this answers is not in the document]
jiminwen-msft: For users not using push notifications, the app will not crash. The crash happens when startPushNotification() API is invoked. And when startPushNotification(), an exception is thrown to let them follow the quick-start guide. I will update quick-start in another PR later. https://github.com/Azure/azure-sdk-for-android/pull/1123/files#diff-4c6963a7914f87efece4f36db9c51cdf12a7f7893011d1609f05c2218e16ad7cR39

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/KeyCreationTimeStore.java
```diff
@@ -0,0 +1,105 @@
...
+        } catch (JsonProcessingException e) {
+            clientLogger.logExceptionAsError(new RuntimeException("Failed to generate JSON object", e));
+        }
+        try (FileOutputStream fos = new FileOutputStream(outputFile);
+            BufferedOutputStream bos = new BufferedOutputStream(fos)) {
+            //convert string to byte array
+            byte[] bytes = jsonStr.getBytes(java.nio.charset.StandardCharsets.UTF_8);
+            //write byte array to file
+            bos.write(bytes);
+            bos.close();
```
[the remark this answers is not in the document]
jiminwen-msft: The try-with-resources statement ensures that each resource is closed at the end of the statement. Any object that implements java.lang.AutoCloseable, which includes all objects which implement java.io.Closeable, can be used as a resource. So actually the bos.clost() could be removed.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -56,6 +82,10 @@ public static RegistrationKeyManager instance() {
         return registrationKeyManager;
     }
 
+    private char[] getKeyStorePassword() {
+        return "com.azure.android.communication.chat.android".toCharArray();
```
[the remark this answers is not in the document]
jiminwen-msft: The password field is actually optional. I am adding it here to slightly increase the difficulty of hack the system. Having a password that keeps changing is challenging to implement. Which way do you suggest?  1. Adding some transform function to make this password complex 2. Skipping the password 3. Alternative ways that I am not aware of.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
jiminwen-msft: This one is a duplicate. I am going to remove it

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -64,84 +94,266 @@ public boolean getLastExecutionSucceeded() {
...
+            RuntimeException runtimeException = new RuntimeException("keyCreationTimeStore failed to delete entry");
+            throw clientLogger.logExceptionAsError(runtimeException);
+        }
+    }
+
+    public Long getCreationTime(String alias) {
+        return keyCreationTimeStore.getCreationTime(alias);
+    }
+
+    public SecretKey getSecreteKey(String alias) {
```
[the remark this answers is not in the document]
jiminwen-msft: Done

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -64,84 +94,266 @@ public boolean getLastExecutionSucceeded() {
...
-            long insertionTime = getInsertionTime(entry);
+    private String secretKeyToStr(SecretKey secretKey) {
+        return Base64Util.encodeToString(secretKey.getEncoded());
+    }
+
+    private int extractIndex(String alias) {
+        if (alias.indexOf(CRYPTO_KEY_PREFIX) == 0) {
+            return Integer.parseInt(alias.replaceAll(CRYPTO_KEY_PREFIX, ""));
+        }
+        return Integer.parseInt(alias.replaceAll(AUTH_KEY_PREFIX, ""));
```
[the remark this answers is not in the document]
jiminwen-msft: Done

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -64,84 +94,266 @@ public boolean getLastExecutionSucceeded() {
...
         }
+
+        //Rotate to fill the empty entries. Move remained entries to lowest index
+        int toIndex = 0;
+        for (int fromIndex : set) {
+            String fromCryptoAlias = CRYPTO_KEY_PREFIX + fromIndex;
+            String fromAuthAlias = AUTH_KEY_PREFIX + fromIndex;
+
+            //This record is cleared or no need to move
+            if (anyEntryMissed(fromCryptoAlias, fromAuthAlias) || toIndex == fromIndex) {
```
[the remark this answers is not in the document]
jiminwen-msft: Removed this checking

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
jiminwen-msft: There is a null checking before converting entry to secretKey.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/KeyCreationTimeStore.java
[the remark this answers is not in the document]
vcolin7: Oh you're right, I don't know how I missed that. Removing both closing statements should be fine.

vcolin7 (commented):
Added a suggestion for the new exception added for when the push notification configuration is incorrect.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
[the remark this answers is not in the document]
jiminwen-msft: Based on offline discussion. The password field will be empty/null. I am investigating changing keystore type from default "AES" to "AndroidKeyStore"

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/KeyMetaDataStore.java
```diff
@@ -0,0 +1,151 @@
...
+                clientLogger.verbose("new file created for storing push notification credentials");
+            }
+        } catch (IOException e) {
+            clientLogger.logExceptionAsError(new RuntimeException("Failed to create key store file", e));
+        }
+        String jsonStr = "";
+        try {
+            jsonStr = new ObjectMapper().writeValueAsString(map);
+        } catch (JsonProcessingException e) {
+            clientLogger.logExceptionAsError(new RuntimeException("Failed to generate JSON object", e));
```
vcolin7, line 83: I feel like we should throw in all these cases.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/KeyMetaDataStore.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+        try (FileOutputStream fos = new FileOutputStream(outputFile);
+            BufferedOutputStream bos = new BufferedOutputStream(fos)) {
+            //convert string to byte array
+            byte[] bytes = jsonStr.getBytes(java.nio.charset.StandardCharsets.UTF_8);
+            //write byte array to file
+            bos.write(bytes);
+        } catch (IOException e) {
+            clientLogger.logExceptionAsError(new RuntimeException("Filed writing key map to file", e));
```
vcolin7, line 92: Same feeling about us throwing.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/KeyMetaDataStore.java
```diff
@@ -0,0 +1,151 @@
...
+    //Write in-memory map into file with Json format
+    private void writeJsonToFile(String path) {
+        File outputFile = new File(path);
+        try {
+            boolean newFile = outputFile.createNewFile();
+            if (newFile) {
+                clientLogger.verbose("new file created for storing push notification credentials");
+            }
+        } catch (IOException e) {
+            clientLogger.logExceptionAsError(new RuntimeException("Failed to create key store file", e));
```
vcolin7, line 77: If we reach this catch block i assume one reason could be because the file already exists, right? What about other scenarios? Is there a case where the file does not exist and could not be created? (e.g. having an invalid path).

I ask so that we don't carry on with the logic below if there is no file available. I guess a catch block further down could handle that scenario...

config/spotbugs/spotbugs-exclude.xml
```diff
@@ -13,10 +13,34 @@
...
+    <Bug pattern="DC_DOUBLECHECK"/>
+  </Match>
+
+  <!-- Over-kill to create static class for typeReference -->
+  <Match>
+    <Class name="com.azure.android.communication.chat.implementation.notifications.fcm.KeyMetaDataStore" />
+    <Bug pattern="SIC_INNER_SHOULD_BE_STATIC_ANON" />
+  </Match>
+
+  <Match>
```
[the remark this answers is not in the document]
jiminwen-msft: Removing this

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,435 @@
...
+                throw clientLogger.logExceptionAsError(new RuntimeException("Failed to decrypt data", e));
+            } catch (UnsupportedEncodingException e) {
+                throw clientLogger.logExceptionAsError(new RuntimeException("Failed to decrypt data", e));
+            } catch (IllegalBlockSizeException e) {
+                throw clientLogger.logExceptionAsError(new RuntimeException("Failed to decrypt data", e));
+            } catch (BadPaddingException e) {
+                throw clientLogger.logExceptionAsError(new RuntimeException("Failed to decrypt data", e));
+            } catch (InvalidKeyException e) {
+                throw clientLogger.logExceptionAsError(new RuntimeException("Failed to decrypt data", e));
+            }
```
vcolin7, line 432: nit; We can collapse all these into a single `catch` block.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,435 @@
...
+        }
+    }
+
+    class DeCryptor {
+        private static final String TRANSFORMATION_SYMMETRIC = "AES/GCM/NoPadding";
+
+        DeCryptor() {
+        }
+
+        String decryptData(final Key decryptionKey, final byte[] encryptedData, final byte[] encryptionIv) {
```
vcolin7, line 413: nit; In cryptography, encrypted data is typically called `ciphertext`.
```suggestion
        String decryptData(final Key decryptionKey, final byte[] ciphertext, final byte[] encryptionIv) {
```

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,435 @@
...
+    private ClientLogger clientLogger = new ClientLogger(RegistrationKeyManager.class);
+
+    private RegistrationKeyManager()  {
+        enCryptor = new EnCryptor();
+        deCryptor = new DeCryptor();
+        try {
+            this.secretKeyGenerator = KeyGenerator.getInstance("AES");
+            this.secretKeyGenerator.init(256);
+            androidKeyGenerator = KeyGenerator.getInstance(
+                KeyProperties.KEY_ALGORITHM_AES, "AndroidKeyStore");
```
vcolin7, line 86: I noticed we are using a secondary `KeyGenerator` with a different implementation than the one by "AndroidKeyStore". From what I can tell below, it's so that we can access its contents via `getEncoded()`. Is this so that the key's entire contents can be sent to the service?

jiminwen-msft: Yes. We could directly send the keys generated by secretKeyGenerator to registration service. The content should be the same as what decrypted from the file. I implement it in a way to send the decrypted content, for easier debugging. It makes it easier to validate the validation of encryption/decryption process. I removed the logging logic in PR to avoid leaking the content on client side.

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/RegistrationKeyManager.java
```diff
@@ -0,0 +1,435 @@
...
+        androidKeyGenerator.generateKey();
+        try {
+            enCryptor.encryptText((SecretKey) keyStore.getKey(alias, null), secretKeyToStr(secretKey));
+        } catch (KeyStoreException e) {
+            e.printStackTrace();
+        } catch (NoSuchAlgorithmException e) {
+            e.printStackTrace();
+        } catch (UnrecoverableKeyException e) {
+            e.printStackTrace();
+        }
```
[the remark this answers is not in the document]
jiminwen-msft: Changing to collapse and throw an exception

sdk/communication/azure-communication-chat/src/main/java/com/azure/android/communication/chat/implementation/notifications/fcm/KeyMetaDataStore.java
jiminwen-msft: file.createNewFile() operation returns false instead of throwing an exception if file already exist. So we just need handle IOException for exception during creating of a file. But I will change to throw an exception in the catch.

vcolin7 (approved):

PR #1178, opened 2022-08-20
Android releasing

jiminwen-msft:
Android release for push notification fix: 
https://github.com/Azure/azure-sdk-for-android/pull/1123/files

### DIFF
--- a/sdk/communication/azure-communication-chat/CHANGELOG.md
+++ b/sdk/communication/azure-communication-chat/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Release History
 ## 2.0.0 (2022-08-26)
 ### Bugs Fixed
-- Push notification registration renewal not guaranteed to happen when Contoso APP is idle or closed.
+- Push notification registration renewal is now guaranteed to happen when an app is idle or closed.
 
 ### Breaking Changes
 - Introduced WorkerManager library which breaks push notification for existing Contoso APPs. (Check https://docs.microsoft.com/azure/communication-services/tutorials/chat-android-push-notification)
--- a/sdk/communication/azure-communication-chat/CHANGELOG.md
+++ b/sdk/communication/azure-communication-chat/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Release History
-## 2.0.0 (2022-08-26)
+## 2.0.0 (2022-08-29)
 ### Features Added
 - Create new `startPushNotifications` API without errorHandler as a parameter.
 
@@ -8,10 +8,10 @@
 
 ### Breaking Changes
 - Introduced the use of WorkManager, which now requires additional configuration in apps that were already using the push notification feature. Check the [push notifications documentation](https://docs.microsoft.com/azure/communication-services/tutorials/chat-android-push-notification) for more information.
+- Increase minSDKVersion from 21 to 23.
 
 ### Other Changes
 - Implement internal key management library to securely persist all the registration secret keys which are not expired.
-- Increase minSDKVersion from 21 to 23.
 - Reduce the expiration time of each push notification registration from 25 hours to 45 minutes.
 - Add `@Deprecated` annotation to `startPushNotifications` API with `errorHandler` as a parameter.
 
--- a/sdk/communication/azure-communication-chat/CHANGELOG.md
+++ b/sdk/communication/azure-communication-chat/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Release History
 ## 2.0.0 (2022-08-26)
+### Features Added
+- Create new `startPushNotifications` API without errorHandler as a parameter.
+
 ### Bugs Fixed
 - Push notification registration renewal is now guaranteed to happen when an app is idle or closed.
 
@@ -11,7 +14,6 @@
 - Increase minSDKVersion from 21 to 23.
 - Reduce the expiration time of each push notification registration from 25 hours to 45 minutes.
 - Add `@Deprecated` annotation to `startPushNotifications` API with `errorHandler` as a parameter.
-- Create new API errorHandler without errorHandler as a parameter.
 
 ## 1.1.0 (2022-04-11)
 ### Features Added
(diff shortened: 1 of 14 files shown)


vcolin7 (commented):
What's the reasoning behind doing a major version bump? Is it only due to the `minSdkLevel` change or are there other things at play? I remember we had some APIs we were planning on deprecating, but I don't remember ever seeing those changes get into the `main` branch.

I was under the impression this kind of version upgrades would happen if we had breaking changes to make on the API level, but there might be other reasons I haven't considered. What is the guidance we follow on this, @JonathanGiles?

jiminwen-msft:
> What's the reasoning behind doing a major version bump? Is it only due to the `minSdkLevel` change or are there other things at play? I remember we had some APIs we were planning on deprecating, but I don't remember ever seeing those changes get into the `main` branch.
> 
> I was under the impression this kind of version upgrades would happen if we had breaking changes to make on the API level, but there might be other reasons I haven't considered. What is the guidance we follow on this, @JonathanGiles?

I am not certain on the version as well actually. This fix will break existing push notification users, though the number is pretty low, less than 100 probably. And there is a bump in min version. Please provide guidance on whether a major version bump is needed. Thanks

jiminwen-msft:
> > What's the reasoning behind doing a major version bump? Is it only due to the `minSdkLevel` change or are there other things at play? I remember we had some APIs we were planning on deprecating, but I don't remember ever seeing those changes get into the `main` branch.
> > I was under the impression this kind of version upgrades would happen if we had breaking changes to make on the API level, but there might be other reasons I haven't considered. What is the guidance we follow on this, @JonathanGiles?
> 
> I am not certain on the version as well actually. This fix will break existing push notification users, though the number is pretty low, less than 100 probably. And there is a bump in min version. Please provide guidance on whether a major version bump is needed. Thanks

The deprecating API change is not included this time. That is an optional change with low priority.

JonathanGiles:
We follow [semantic versioning](https://semver.org/) in Java and Android. We will need to discuss in more depth what is breaking, why it is necessary, and how we have mitigated the breaks for existing users (e.g. updates to documentation, samples, release notes, etc) before we can determine what the version increment should be.

sdk/communication/azure-communication-chat/CHANGELOG.md
```diff
@@ -1,13 +1,17 @@
...
-
-### Features Added
+## 2.0.0 (2022-08-26)
+### Bugs Fixed
+- Push notification registration renewal not guaranteed to happen when Contoso APP is idle or closed.
 
 ### Breaking Changes
-
-### Bugs Fixed
+- Introduced WorkerManager library which breaks push notification for existing Contoso APPs. (Check https://docs.microsoft.com/azure/communication-services/tutorials/chat-android-push-notification)
```
vcolin7, line 7:
```suggestion
- Introduced the use of WorkManager, which now requires additional configuration in apps that were already using the push notification feature. Check the [push notifications documentation](https://docs.microsoft.com/azure/communication-services/tutorials/chat-android-push-notification) for more information.
```

sdk/communication/azure-communication-chat/CHANGELOG.md
```diff
@@ -1,13 +1,17 @@
 # Release History
-## 1.2.0-beta.1 (Unreleased)
-
-### Features Added
+## 2.0.0 (2022-08-26)
+### Bugs Fixed
+- Push notification registration renewal not guaranteed to happen when Contoso APP is idle or closed.
```
vcolin7, line 4: Are we saying this is no longer happening? The wording is confusing and could make people think the renewal is not guaranteed to happen when in fact, it is.
```suggestion
- Push notification registration renewal is now guaranteed to happen when an app is idle or closed.
```

sdk/communication/azure-communication-chat/CHANGELOG.md
```diff
@@ -1,13 +1,17 @@
...
 
 ### Breaking Changes
-
-### Bugs Fixed
+- Introduced WorkerManager library which breaks push notification for existing Contoso APPs. (Check https://docs.microsoft.com/azure/communication-services/tutorials/chat-android-push-notification)
 
 ### Other Changes
+- Implement internal key management library to securely persist all the registration secret keys which are not expired.
+- Increase minSDKVersion from 21 to 23.
```
vcolin7, line 11: Not sure if this should be under the breaking changes section, as it can technically be "reversed". Any thoughts @JonathanGiles?

sdk/communication/azure-communication-chat/CHANGELOG.md
```diff
@@ -1,13 +1,17 @@
...
+- Reduce the expiration time of each push notification registration from 25 hours to 45 minutes.
+- Add deprecate annotation to startPushNotifications API with errorHandler as a parameter.
```
vcolin7, line 13:
```suggestion
- Add `@Deprecated` annotation to `startPushNotifications` API with `errorHandler` as a parameter.
```

sdk/communication/azure-communication-chat/CHANGELOG.md
```diff
@@ -1,13 +1,17 @@
...
+- Create new API errorHandler without errorHandler as a parameter.
```
vcolin7, line 14: We should probably put this one under `### Features Added` like we've done with other new APIs.

sdk/communication/azure-communication-chat/CHANGELOG.md
jiminwen-msft: I think "Increase minSDKVersion from 21 to 23" could be put under breaking changes. Because it is breaking change in this release (2.0.0). When we do a reverse, it would be done in another release like 2.1.0. 

Additionally, put it under breaking change section is more obvious for users. They are likely to ignore the "Other changes" section.

vcolin7: Sounds good, let's do it then.

vcolin7 (approved):

PR #1183, opened 2022-08-29
Updated Windows VM image versions in `archetype-android-release.yml`.

vcolin7:


weshaggard (approved):

JimSuplizio (approved):
As per our conversation, these were taking from the Java repository for the same/similar jobs.

PR #1236, opened 2022-11-11
[BOT RULES] Communication - Common

ronniegeraghty:
Adding bot rules to match changes in code owners in #1234


jsquire:
Thanks for putting this together, but let's leave the current "Communication" labeling.   We don't typically want to tag PRs at this level of granularity.   😄

PR #1240, opened 2022-11-15
Update Android to use ESRP publishing

### LINKED ISSUE #1169: Use ESRP tasks in ADO to release Android maven packages

JimSuplizio:
This is the initial set of changes to get Android publishing through ESRP.

1. First, move the Resource->Repositories from the individual ci.yml files into the archetype-sdk-client.yml. This is what we do in Java for the azure-sdk-build-tools repository so we don't have to add it to every ci.yml file. This also required changing the **stages** to **extends** in the ci.yml files.
2. Add a ci.yml file for template so it can actually be built and tested. Template is the library used for release testing.
3. Fixed template so it'll actually work. Note: The updates to template's gradle.properties was done for release purposes.
4. Like Java, move the tooling into the Android repository. Right now, the files are copied from the Java repository to just get things working and the set of files to do this is rather small. 

- eng/pipelines/templates/steps/java-publishing.yml
- eng/pipelines/templates/steps/publish-artifact.yml
- eng/scripts/MavenPackaging.ps1
- eng/scripts/Publish-MavenPackages.ps1
- eng/scripts/SignAndHash-MavenPackages.ps1
- eng/maven.publish.settings.xml

5. Bring forward githubIODocs publishing, which was using an archaic tools script. This required adding the Publish-andriod-GithubIODocs to the LanguageSettings. 

Fixes #1169


JimSuplizio:
@weshaggard @vcolin7 - Here's the deal, in order to move the azure-sdk-build-tools ref forward, to what Java is using, the doc publishing had to change. The previous ref was like a year old and was stuck there because it was using the docs publishing pre-azure storage and the files it was using have been gone for like a year. Unfortunately, this means that this respository's docs githubIO needs to be converted to use the blob storage, just like everything else as no new docs will be published to the [original location](https://github.com/Azure/azure-sdk-for-android/tree/gh-pages) @sima-zhu is going to be writing a transition doc.

More than half the commits here are from dealing with the githubIO publishing. The scripts are pretty unforgiving with missing functions and sometimes let you know and sometimes fail silently. A number of commits were made to add more temporary diagnostics output and those diagnostics changes have been removed. The final commit is a version update for template.

@vcolin7 with these changes it should be trivial to add nightly publishing to the android dev feed but it's going to require something that can manipulate the versions in the repository.

vcolin7 (approved):

sdk/template/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+    include:
+      - main
+      - feature/*
+      - hotfix/*
+      - release/*
+  paths:
+    include:
+      - sdk/core/
+      - eng/
+      - build.gradle
```
weshaggard, line 14: Lets add the template as a trigger path.

eng/pipelines/templates/steps/publish-artifact.yml
```diff
@@ -0,0 +1,27 @@
+# This step is used to prevent duplication of artifact publishes when there is an issue that would prevent the overall success of the job.
+# Ensuring that we only publish when successful (and two a differently named artifact otherwise) will allow easy retry on a build pipeline
```
weshaggard, line 2: This is already in eng/common.

weshaggard (approved):

eng/pipelines/templates/steps/publish-artifact.yml
JimSuplizio: Removed as per our discussion. It was incorrectly added.

sdk/template/ci.yml
JimSuplizio: changed from sdk/core to sdk/template

PR #1307, opened 2023-02-21
[Android] Added support for MicrosoftBotIdentifier

AikoBB:
# Description

Introducing a new type of CommunicationIdentifier - MicrosoftBotIdentifier and the logic is covered with tests.
Implementation of the [User Story 3133111](https://skype.visualstudio.com/SPOOL/_workitems/edit/3133111): [SDK][Common][Mobile] As a developer I can identify if a user is a Teams User or Bot)

API view - [azure-communication-common](https://apiview.dev/Assemblies/Review/c8145aaf77504830b9faf0809da1018f)

### DIFF
--- a/sdk/communication/azure-communication-common/CHANGELOG.md
+++ b/sdk/communication/azure-communication-common/CHANGELOG.md
@@ -1,11 +1,12 @@
 # Release History
 
-## 1.2.0-beta.1 (Unreleased)
+## 2.0.0-beta.1 (Unreleased)
 
 ### Features Added
 - Added support for a new communication identifier `MicrosoftBotIdentifier`.
 
 ### Breaking Changes
+- Introduction of `MicrosoftBotIdentifier` is a breaking change. It will affect code that relied on using `UnknownIdentifier` with a rawID starting with `28:`
 
 ### Bugs Fixed
 
--- a/sdk/communication/azure-communication-common/CHANGELOG.md
+++ b/sdk/communication/azure-communication-common/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.2.0-beta.1 (Unreleased)
 
 ### Features Added
+- Added support for a new communication identifier `MicrosoftBotIdentifier`.
 
 ### Breaking Changes
 
--- a/sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationIdentifier.java
+++ b/sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationIdentifier.java
@@ -6,6 +6,20 @@ package com.azure.android.communication.common;
  * Common communication identifier for Communication Services
  */
 public abstract class CommunicationIdentifier {
+    private static final String ACS_USER_PREFIX = "8:acs:";
+    private static final String ACS_GCCH_USER_PREFIX = "8:gcch-acs:";
+    private static final String ACS_DOD_USER_PREFIX = "8:dod-acs:";
+    private static final String SPOOL_USER_PREFIX = "8:spool:";
+    protected static final String TEAMS_PUBLIC_USER_PREFIX = "8:orgid:";
+    protected static final String TEAMS_GCCH_USER_PREFIX = "8:gcch:";
+    protected static final String TEAMS_DOD_USER_PREFIX = "8:dod:";
+    protected static final String TEAMS_ANONYMOUS_USER_PREFIX = "8:teamsvisitor:";
+    private static final String BOT_PUBLIC_PREFIX = "28:orgid:";
+    private static final String BOT_DOD_PREFIX = "28:dod:";
+    private static final String BOT_DOD_GLOBAL_PREFIX = "28:dod-global:";
+    private static final String BOT_GCCH_PREFIX = "28:gcch";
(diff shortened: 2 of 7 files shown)


sdk/communication/azure-communication-common/src/test/java/com/azure/android/communication/common/CommunicationIdentifierTests.java
```diff
@@ -113,12 +211,38 @@ public void createIdentifierFromRawId()
         assertIdentifier("4:207ffef6-9444-41fb-92ab-20eacaae2768", new PhoneNumberIdentifier("207ffef6-9444-41fb-92ab-20eacaae2768"));
         assertIdentifier("4:207ffef6-9444-41fb-92ab-20eacaae2768_207ffef6-9444-41fb-92ab-20eacaae2768", new PhoneNumberIdentifier("207ffef6-9444-41fb-92ab-20eacaae2768_207ffef6-9444-41fb-92ab-20eacaae2768"));
         assertIdentifier("4:+112345556789_207ffef6-9444-41fb-92ab-20eacaae2768", new PhoneNumberIdentifier("+112345556789_207ffef6-9444-41fb-92ab-20eacaae2768"));
-        assertIdentifier("28:45ab2481-1c1c-4005-be24-0ffb879b1130", new UnknownIdentifier("28:45ab2481-1c1c-4005-be24-0ffb879b1130"));
+
+        assertIdentifier(CommunicationIdentifier.BOT_PREFIX + microsoftBotId, new MicrosoftBotIdentifier(microsoftBotId, true));
```
petrsvihlik, line 215: same as in ios - use hardcoded fixtures

petrsvihlik (changes requested):

petrsvihlik (approved):

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/MicrosoftBotIdentifier.java
```diff
@@ -0,0 +1,141 @@
...
+            return true;
+        }
+
+        if (!(that instanceof MicrosoftBotIdentifier)) {
+            return false;
+        }
+
+        MicrosoftBotIdentifier thatId = (MicrosoftBotIdentifier) that;
+
+        if (cloudEnvironment != null && !cloudEnvironment.equals(thatId.cloudEnvironment)) {
```
AikoBB, line 101: will update to check only for the raw id

sdk/communication/azure-communication-common/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 1.2.0-beta.1 (Unreleased)
 
 ### Features Added
+- Added support for a new communication identifier `MicrosoftBotIdentifier`.
```
petrsvihlik, line 6: I'd add a record under both features added and breaking changes...see my comment in the js sdk

petrsvihlik (changes requested):

ostoliarova-msft (approved):

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationIdentifier.java
```diff
@@ -6,6 +6,21 @@
  * Common communication identifier for Communication Services
  */
 public abstract class CommunicationIdentifier {
+    static final String ACS_USER_PREFIX = "8:acs:";
+    static final String ACS_USER_GCCH_CLOUD_PREFIX = "8:gcch-acs:";
+    static final String ACS_USER_DOD_CLOUD_PREFIX = "8:dod-acs:";
+    static final String SPOOL_USER_PREFIX = "8:spool:";
```
petrsvihlik, line 12: interesting...we don't have this in other langs

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/MicrosoftBotIdentifier.java
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * Set full id of the identifier.
+     * RawId is the encoded format for identifiers to store in databases or as stable keys in general.
+     *
+     * @param rawId full id of the identifier.
+     * @return MicrosoftBotIdentifier object itself.
+     */
+    @Override
+    public MicrosoftBotIdentifier setRawId(String rawId) {
```
petrsvihlik, line 89: we could possibly avoid the whole `rawIdSet` fiddling if we didn't expose the `setRawId` method publicly. do we need to do that?

petrsvihlik (approved):

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/MicrosoftBotIdentifier.java
AikoBB: so as the setRawId is available for the public already from the parent class of the bot identifier, so I think we can not avoid from it

sdk/communication/azure-communication-common/src/main/java/com/azure/android/communication/common/CommunicationIdentifier.java
petrsvihlik: never noticed that :)

PR #1341, opened 2023-04-04
[android] azure-communication-common release

AikoBB:
azure-communication-common a beta release - 2.0.0-beta.1

- added a new identifier `MicrosoftBotIdentifier
- Introduction of `MicrosoftBotIdentifier` is a breaking change

### DIFF
--- a/sdk/communication/azure-communication-common/CHANGELOG.md
+++ b/sdk/communication/azure-communication-common/CHANGELOG.md
@@ -1,16 +1,12 @@
 # Release History
 
-## 2.0.0-beta.1 (Unreleased)
+## 2.0.0-beta.1 (2023-04-04)
 
 ### Features Added
 - Added support for a new communication identifier `MicrosoftBotIdentifier`.
 
 ### Breaking Changes
 - Introduction of `MicrosoftBotIdentifier` is a breaking change. It will affect code that relied on using `UnknownIdentifier` with a rawID starting with `28:`
-
-### Bugs Fixed
-
-### Other Changes
 
 ## 1.1.0 (2022-11-11)
 
--- a/sdk/communication/azure-communication-common/README.md
+++ b/sdk/communication/azure-communication-common/README.md
@@ -20,7 +20,7 @@ This package contains common code for Azure Communication Service libraries.
   APIs that would require the Java 8+ API desugaring provided by Android Gradle plugin 4.0.0.
 
 ### Versions available
-The current version of this library is **1.1.0**.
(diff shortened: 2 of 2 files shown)


mjafferi-msft (approved):

maximrytych-ms (approved):

sdk/communication/azure-communication-common/CHANGELOG.md
```diff
@@ -1,17 +1,13 @@
 # Release History
 
-## 2.0.0-beta.1 (Unreleased)
+## 2.0.0-beta.1 (2023-04-04)
 
 ### Features Added
 - Added support for a new communication identifier `MicrosoftBotIdentifier`.
 
 ### Breaking Changes
 - Introduction of `MicrosoftBotIdentifier` is a breaking change. It will affect code that relied on using `UnknownIdentifier` with a rawID starting with `28:`
```
vcolin7, line 9: I talked to the folks from iOS and they said they did not consider this behavioral breaking change to warrant a major version update, just like Python is not. I am also curious why there was a difference between those languages and C# regarding this decision. What did the Java architects say about it?

AikoBB: Yes, we had a discussion around that today for iOS. For Java[(API view)](https://apiview.dev/Assemblies/Review/d70b841f485d4854870aa13688b23d2e?diffRevisionId=51937aef84e74c89b6bc685432544205&diffOnly=False&revisionId=527402c5099b4112b0559349a22548e3&doc=False) and android[(API view)](https://apiview.dev/Assemblies/Review/c8145aaf77504830b9faf0809da1018f?diffRevisionId=1a4bf12d8b174f7d83c6a0929fb78114&diffOnly=False&revisionId=ea917d18a0744b4ab411823a1b070007&doc=False) we got approval for the API changes where were the version change was included. We double checked in the e-mail whether it should be minor version bump and got feedback that from C# architects that it should be major version update as it's a behavioral breaking change in a shared library, and only got feedback from Python architects to go with a minor version bump. So we shipped with a major version update for java, js, c#

vcolin7: We should ask the Java architects just to make sure as early as we can, we want to avoid having to change the version from 2.0.0 to some minor update when the time to release as GA comes.

vcolin7 (approved):
@AikoBB I talked to our architects and, since we already release version `2.0.0-beta.1`, it would be a bad idea to make it a GA for a lower, minor version. So we'll keep things like that for Java and Android will follow suit.

PR #1478, opened 2023-12-01
[DO NOT MERGE] Commit API docs

dipidoo:
To keep repo history small, always amend and force-push this commit rather than adding commits to the gh-pages branch.


dipidoo:
This PR is meant only to show the staged docs and not meant to be completed--it is meant to be completed via commit amend and force push

.DS_Store
vcolin7, line 1: We should exclude these `.DS_Store` files using `.gitignore`, they're Mac OS-specific and serve no purpose to the repo.

Issue #1483, opened 2024-01-04
[BUG] Duplicate Chat Events Received for the same event


shuchenku:
**Describe the bug**
For each chat event, `ChatAsyncClient's` event handler receives two `ChatEvent` instances. This causes our client to display the events twice in the chat thread.
Sometimes the payloads are the exact same, while other times the payloads are different only in the `recipient` and `sender` attributes (same Id but different objects).

***Exception or Stack Trace***
Message # 1 - exact same payload:
- 1st message
```
 payload = {ChatMessageReceivedEvent@29016} 
 acsChatMessageMetadata = "null"
 recipientId = "acs:fc9b024a-9263-4319-b8e6-2a613c1189a7_00000017-c429-2b60-df68-563a0d00cc1e"
 content = "hahahha"
 createdOn = {OffsetDateTime@29019} 
  dateTime = {LocalDateTime@29032} 
  offset = {ZoneOffset@29033} 
  shadow$_klass_ = {Class@27514} 
  shadow$_monitor_ = 0
 id = "1704391395508"
 messageType = "Text"
 metadata = null
 priority = ""
 senderDisplayName = "******"
 type = {ChatMessageType@29024} 
 version = "1704391395508"
 recipient = {CommunicationUserIdentifier@29026} 
  id = "8:acs:fc9b024a-9263-4319-b8e6-2a613c1189a7_00000017-c429-2b60-df68-563a0d00cc1e"
  rawId = "8:acs:fc9b024a-9263-4319-b8e6-2a613c1189a7_00000017-c429-2b60-df68-563a0d00cc1e"
  shadow$_klass_ = {Class@28832} 
  shadow$_monitor_ = 0
 recipientMri = "8:acs:fc9b024a-9263-4319-b8e6-2a613c1189a7_00000017-c429-2b60-df68-563a0d00cc1e"
 sender = {CommunicationUserIdentifier@29029} 
  id = "8:acs:fc9b024a-9263-4319-b8e6-2a613c1189a7_00000019-6093-d151-655d-573a0d0010e1"
  rawId = "8:acs:fc9b024a-9263-4319-b8e6-2a613c1189a7_00000019-6093-d151-655d-573a0d0010e1"
  shadow$_klass_ = {Class@28832} 
  shadow$_monitor_ = 0
 senderId = "8:acs:fc9b024a-9263-4319-b8e6-2a613c1189a7_00000019-6093-d151-655d-573a0d0010e1"
[...]

PR #1491, opened 2024-01-16
Fix duplicate real time notifications

jiminwen-msft:
Fixes: #1483.

https://portal.microsofticm.com/imp/v3/incidents/details/457409680/home.

Customers report getting two real time notifications for single request using Android SDK. We are able to reproduce the issue locally. The root cause is the outdated Trouter template which misses the Trouter-control header. Directly modifying the template is risky because most users use this outdated template. 

Making this change to use latest version of template "AcsAndroid_Chat_test_1.7". Tested this change locally

### DIFF
--- a/sdk/communication/azure-communication-chat/CHANGELOG.md
+++ b/sdk/communication/azure-communication-chat/CHANGELOG.md
@@ -2,7 +2,7 @@
 ## 2.1.0 (2024-01-18)
 
 #### Bugs Fixed
-Address the issue described in https://github.com/Azure/azure-sdk-for-android/issues/1483. The SDK now incorporates more recent Trouter template to prevent Android devices from receiving repeated real-time notifications.
+- Address the issue described in https://github.com/Azure/azure-sdk-for-android/issues/1483. The SDK now incorporates a more recent Trouter template to prevent Android devices from receiving duplicate real-time notifications.
 
 ## 2.0.1 (2023-10-25)
 
--- a/sdk/communication/azure-communication-chat/CHANGELOG.md
+++ b/sdk/communication/azure-communication-chat/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Release History
-## 2.1.0 (2024-01-18)
+## 2.0.2 (2024-01-18)
 
 #### Bugs Fixed
 - Address the issue described in https://github.com/Azure/azure-sdk-for-android/issues/1483. The SDK now incorporates a more recent Trouter template to prevent Android devices from receiving duplicate real-time notifications.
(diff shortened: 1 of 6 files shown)


sdk/communication/azure-communication-chat/env-prod.properties
```diff
@@ -11,7 +11,7 @@ trouter.registration.hostname="teams.microsoft.com/registrar/prod"
 trouter.registration.hostname.dod="registrar.dod.teams.microsoft.us"
 trouter.registration.hostname.gcch="registrar.gov.teams.microsoft.us"
 trouter.application.id="AcsAndroid"
-trouter.template.key="AcsAndroid_Chat_1.3"
+trouter.template.key="AcsAndroid_Chat_1.7"
```
jpeng-ms, line 14: one question, it seems like Ali has added attachments to 1.7 which we don't plan to support right now, can you find a template version that doesn't have this attachments please?

jpeng-ms: maybe 1.5? would you mind testing it and see if everything is working fine with 1.5 please?

jpeng-ms (approved):

sdk/communication/azure-communication-chat/env-prod.properties
jiminwen-msft: Changed to 1.5 and tested locally

jpeng-ms (approved):
looks good thanks!

sdk/communication/azure-communication-chat/CHANGELOG.md
```diff
@@ -1,13 +1,8 @@
...
 
-### Features Added
-
-### Breaking Changes
-
-### Bugs Fixed
-
-### Other Changes
+#### Bugs Fixed
+Address the issue described in https://github.com/Azure/azure-sdk-for-android/issues/1483. The SDK now incorporates more recent Trouter template to prevent Android devices from receiving repeated real-time notifications.
```
vcolin7, line 5: nit;
```suggestion
- Address the issue described in https://github.com/Azure/azure-sdk-for-android/issues/1483. The SDK now incorporates a more recent Trouter template to prevent Android devices from receiving duplicate real-time notifications.
```

vcolin7 (commented):
Things look good on my end. I'm just curious about why we're doing a minor version bump (`2.0.1` -> `2.1.0`) as opposed to a patch (`2.0.1` -> `2.0.2`). Are we exposing any new from the Trouter library?

jiminwen-msft:
> Things look good on my end. I'm just curious about why we're doing a minor version bump (`2.0.1` -> `2.1.0`) as opposed to a patch (`2.0.1` -> `2.0.2`). Are we exposing any new from the Trouter library?

PR #1508, opened 2024-02-03
[DO NOT MERGE] Commit API docs

dipidoo:
To keep repo history small, always amend and force-push this commit rather than adding commits to the gh-pages branch


vcolin7 (changes requested):
Adding comment to block merging.

vcolin7 (commented):
@dipidoo LGTM!

PR #1576, opened 2024-05-04
Commit API docs

dipidoo:
To keep repo history small, always amend and force-push this commit rather than adding commits to the gh-pages branch


vcolin7:
I checked out your branch and compared it against main locally (GH was not kind to this PR). The changes look good. I'll force-push them to the branch as usual.

PR #1722, opened 2024-10-24
Jimin/codeowner

jiminwen-msft:
1. Update owner information. Revert to Azure/azure-sdk-communication-code-reviewers
2. Add default constructor for realtimeConfig. Otherwise, the deserialization in realtimeConfigClient failed, since a default constructor is required.


.github/CODEOWNERS
```diff
@@ -22,7 +22,7 @@
 # Service teams
 
 # PRLabel: %Communication
-/sdk/communication/                                 @vcolin7 @Azure/azure-sdk-write-communication
+/sdk/communication/                                 @vcolin7 @Azure/azure-sdk-communication-code-reviewers
 
 # PRLabel: %Communication - Common
 /sdk/communication/azure-communication-common/      @Azure/acs-identity-sdk @AikoBB @maximrytych-ms @mjafferi-msft
```
JoshuaLai, line 28: just a call out it looks like github doesn't recognize this group: `@Azure/acs-identity-sdk` maybe we should just remove it?

AikoBB: Hi @JoshuaLai ! We created this group a few years ago and we can take a look at it why it can't be recognized, and fix it either by removing or having a proper group

.github/CODEOWNERS
```diff
@@ -22,7 +22,7 @@
 # Service teams
 
 # PRLabel: %Communication
-/sdk/communication/                                 @vcolin7 @Azure/azure-sdk-write-communication
+/sdk/communication/                                 @vcolin7 @Azure/azure-sdk-communication-code-reviewers
```
weshaggard, line 25: Why are you reverting this change?

vcolin7 (approved):

PR #1832, opened 2025-03-06
Updates for 1ES ESRP Publishing Changes

JimSuplizio:
Android still releases single library which means the GPG signing step creates directories that are filtered by the artifactId and groupId. This means that, unlike Java, we need to create an ESRP/GPG signed directories for each library at signing. This already happens in the original Release step. The reason why I'm creating a flattened directory is that the original step, when filtered with group/artifact ends up creating a directory with the version part of the directory structure which isn't something I have access to in yml. Flattening the directory allows me to create a "packages-<artifactId>-esrp-flattened" artifact directory which allows the pre-library release to get the correct release directory using the artifactId.

Template's gradle.properties was changed so I could fully test the release.

The cspell change is to exclude all things under eng, like other languages, instead of just eng/common.

The 1es-redirect.yml changes - changing the signing job from a deployment to a job lit up a bunch of extra checks which were apparently never run in Android. The 1ES SpotBugs had to be completely disabled, like Java, because the 1ES injected task doesn't actually run correctly but also the pipeline is running spotbugs in the analyze job.

Note: There are going to need to be more 1ES changes, I'd rather mess with those outside of this PR which is specifically for getting release enabled.

Testing:
[android - communication](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4625985&view=artifacts&type=publishedArtifacts) was used to test creation of the multiple artifact directories.
[android - template](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4626325&view=results) is being used to test the full release which successfully released without issue.


eng/pipelines/templates/stages/archetype-android-release.yml
```diff
@@ -19,36 +19,72 @@ stages:
...
+          name: azsdk-pool-mms-win-2022-general
+          image: azsdk-pool-mms-win-2022-1espt
+          os: windows
+        steps:
+          - checkout: self
+
+          - checkout: azure-sdk-build-tools
+            path: azure-sdk-build-tools
+
+          - template: /eng/common/pipelines/templates/steps/sparse-checkout.yml
```
benbp, line 34: Either this or `checkout: self` should be sufficient I think, together they are redundant.

eng/pipelines/templates/steps/java-dev-feed-publishing.yml
```diff
@@ -0,0 +1,46 @@
...
+  OutputDirectory: not-specified
+  RepositoryUrl: not-specified
+  GroupID:
+  ArtifactID:
+  GPGExecutablePath: $(Pipeline.Workspace)/azure-sdk-build-tools/tools/gpg/bin/gpg.exe
+  JavaRepoRoot: $(Pipeline.Workspace)/azure-sdk-for-java
+  ShouldPublish: true
+
+steps:
+  - ${{if eq(parameters.Target, 'JavaDevFeed')}}:
```
benbp, line 12: This parameter is not specified above

eng/pipelines/templates/stages/archetype-android-release.yml
JimSuplizio: No, unfortunately, they're both necessary. These were lifted directly from the [original publish step](https://github.com/Azure/azure-sdk-for-android/pull/1832/files/22f06481580ab6edc5c9536a0d5840690eefd3f6#diff-2876c69c83479f9fb61da557dcf20fe4049d06f98023a3743b63b2a417fdb626L109).

eng/pipelines/templates/steps/java-dev-feed-publishing.yml
JimSuplizio: Apparently, it wasn't in the original [java-publishing.yml](https://github.com/Azure/azure-sdk-for-android/pull/1832/files/22f06481580ab6edc5c9536a0d5840690eefd3f6#diff-8cf63af57e4d881a5b7c454bf6be70fbc706a37e1af3e70f989f19f5d9adf1c5L1) either. I'm not making updates in here, these step files are copied directly from azure-sdk-for-java.

PR #1833, opened 2025-03-06
1es-redirect changes to deal with warnings

JimSuplizio:
The [1es ESRP publishing changes](https://github.com/Azure/azure-sdk-for-android/pull/1832), after changing Signing to a job from a deployment lit up some 1ES checks that need to be fixed.

1. Created a CredScanSuppression.json file to suppress a warning caused by the test proxy's cert in a different [PR](https://github.com/Azure/azure-sdk-for-android/pull/1834) since the download-credscan-suppressions.yml needs to pull it from there.
2. Remove the superfluous `- checkout: self` (Yes @benbp, you were right!)
3. Turn off CodeQL and add credscan to the 1es-redirect.yml


benbp (approved):

eng/pipelines/templates/steps/download-credscan-suppressions.yml
```diff
@@ -0,0 +1,12 @@
+# 1es PT credscan step will fail if the suppression file path we use in other stages does not exist here.
```
weshaggard, line 1: I'm good with this for now but I really want to eliminate these templates. I know @hallipr is working on a strategy for the .NET release where we don't need this any longer.

weshaggard (approved):

eng/pipelines/templates/steps/download-credscan-suppressions.yml
JimSuplizio: Yeah, @benbp commented when we had to add this that you weren't going to like it. I realize this will have to change both here and Java.

benbp: If we can remove the build tools repo that will solve it, or are we trying another approach?

PR #1834, opened 2025-03-06
Add CredScanSuppression.json

JimSuplizio:
This needs to be added separately from download-credscan-suppressions.yml, which uses it, because if it's not in main already, the step will fail.


JimSuplizio:
There's nothing that'll run which will test this. I need this in main so my other PR which is adding the usage can pass.

PR #1925, opened 2025-06-15
Post release updates for ACS Common

alexandra142:
applied the version increment after the ACS Common release


alexandra142:
@azure-sdk-eng run check-license

PR #2222, opened 2026-05-04
Transition to GitHub App auth, remove azuresdk-github-pat usage

scbedd:
This pull request introduces minor configuration changes to the Azure pipeline YAML files. The main update is the explicit addition of an empty `AuthToken` parameter in two locations, likely to ensure consistent parameterization or to override a default value.

Most important changes:

**Pipeline configuration updates:**

* Added an explicit `AuthToken: ''` parameter to the `prepare-pipelines.yml` pipeline template to ensure the parameter is set, even if empty.
* Set `AuthToken: ''` in the `archetype-android-release.yml` release stage, likely to standardize authentication settings or prevent accidental token usage.

Related to Azure/azure-sdk-tools#9842

### DIFF
--- a/eng/pipelines/prepare-pipelines.yml
+++ b/eng/pipelines/prepare-pipelines.yml
@@ -7,3 +7,4 @@ extends:
   template: /eng/common/pipelines/templates/jobs/prepare-pipelines.yml
   parameters:
     Repository: Azure/azure-sdk-for-android
+    AuthToken: ''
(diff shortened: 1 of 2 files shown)


eng/pipelines/prepare-pipelines.yml
```diff
@@ -7,3 +7,4 @@ extends:
   template: /eng/common/pipelines/templates/jobs/prepare-pipelines.yml
   parameters:
     Repository: Azure/azure-sdk-for-android
+    AuthToken: ''
```
chidozieononiwu:
```suggestion
```

benbp (approved):

